PR #3, opened 2016-02-23
Reorganize newtmgr commands

gordonchaffee:
* Move commands to cmd package in newtmgr/cmd
* Place each command into its own file


sterlinghughes:
Hi Gordon.

Right now your pull request requires source code from the runtime xolo project, which is not a part of the ASF project and is specific to runtime's services.

It looks like this is to leverage the connection profile manager, which is different between runtime and newtmgr (because the connection profile manager in runtime holds both connections to newt devices, and to runtime's cloud services.)  This was abstracted into an interface, so that either connection manager (newtmgr or runtime), could be passed into these functions.  Could you either keep that, or figure out a different way for newtmgr to avoid relying on runtime's libraries here? 

We don't want the ASF project to have special references to our cloud service.  But we do want to make the libraries easy for us (or others) to hook into.

Thanks,
Sterling

gordonchaffee:
goimports was finding an unintended import in the runtimeinc repository, so that was getting pulled in unnecessarily. After discussing some of the intent of the cli directory with Sterling, I've done work to do the following:

- Moved cli/cp.go => config/connprofile.go. This file manages configurations rather than implements commands. To avoid a circular dependency, it needs to be in a separate directory from the commands.
- Putting all the commands in the cli/ directory rather than in cmd

PR #29, opened 2016-12-28
Better handling of Windows paths

simonratner:
Go tries to be smart in how it handles paths, and uses a platform-appropriate separator (backslash on Windows).

However, many Windows tool ports don't deal with backslashes in paths very well, and in some cases like `cygwin` or `msys`, backslash is actually the wrong separator. On the other hand, Windows has natively supported slash-separated paths for a while now, and all tools handle those correctly.

This change normalizes paths used by `newt` a little bit; trying to consistently use slashes for separators (which a lot of code is doing via concatenation anyway), and removing trailing slashes from directories. With this change, `newt` runs quite well on Windows.

### DIFF
--- a/newt/vendor/mynewt.apache.org/newt/util/util.go
+++ b/newt/vendor/mynewt.apache.org/newt/util/util.go
@@ -310,7 +310,12 @@ func ShellCommandLimitDbgOutput(
 	}
 
 	if err != nil {
-		return o, NewNewtError(string(o))
+		log.Debugf("err=%s", err.Error())
+		if len(o) > 0 {
+			return o, NewNewtError(string(o))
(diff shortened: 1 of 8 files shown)


newt/vendor/mynewt.apache.org/newt/util/util.go
```diff
@@ -310,7 +310,12 @@ func ShellCommandLimitDbgOutput(
 	}
 
 	if err != nil {
-		return o, NewNewtError(string(o))
```
simonratner: Go isn't really my thing; should I remove the `/vendor` changes and let some re-vendoring process recreate those, or is it ok to just commit both?

PR #30, opened 2017-01-05
Fix load and debug options due to duplicated '-j' short option

sjanc:
Those two commands also provide short '-j' option for --extrajtagcmd
which is conflicting with '-j' short option for --jobs. Since those
are extra parameters just expect user to use long option for those.


ccollins476ad:
Thanks, Szymon.

PR #32, opened 2017-01-25
Fixes MYNEWT-556

cwanda:
MYNEWT-556 Supports detection of Priority Violation
3 factors contributed to newt ignoring the override silenty.
1) The code only checked for Lateral override
2) The code processed syscfg.defs and syscfg.vals by package priority
   (lowest first) So when a package overrides the setting of higher
   priority the setting definition has not been processed yet, and was treated as
   overriding an undefined setting, which is treated as a warning.
3) newt build did not print out warning message for overriding an undefined setting (See MYNEWT-557)

Fix involves:
1) Process  all the syscfg.defs for all package first. Then process the syscfg.vals for each package.
2) Save the package that defined the setting in CfgEntry
3) Create CfgPriority to save priority violation: setting name, package that defined the setting,
   and package that override the setting.
4) Reordered PackageType definition by package priority (in increasing order of priority). Also reoreded
   PackageTypeNames to match order of PackageType definition (not functionally needed)
5) Compare priority types between the package overriding the value and the package defining the setting to detect prioirty violations. Newt aborts the build.

### DIFF
--- a/newt/pkg/package.go
+++ b/newt/pkg/package.go
@@ -30,28 +30,29 @@ const (
 	PACKAGE_STABILITY_DEV    = "dev"
 )
 
+// Define constants with values of increasing priority.
 const (
-	PACKAGE_TYPE_APP interfaces.PackageType = iota
-	PACKAGE_TYPE_BSP
-	PACKAGE_TYPE_SDK
-	PACKAGE_TYPE_COMPILER
-	PACKAGE_TYPE_LIB
-	PACKAGE_TYPE_TARGET
-	PACKAGE_TYPE_UNITTEST
-	PACKAGE_TYPE_GENERATED
+	PACKAGE_TYPE_COMPILER interfaces.PackageType = iota
 	PACKAGE_TYPE_MFG
+	PACKAGE_TYPE_SDK
+	PACKAGE_TYPE_GENERATED
+	PACKAGE_TYPE_LIB
+	PACKAGE_TYPE_BSP
+	PACKAGE_TYPE_UNITTEST
+	PACKAGE_TYPE_APP
+	PACKAGE_TYPE_TARGET
 )
 
 var PackageTypeNames = map[interfaces.PackageType]string{
-	PACKAGE_TYPE_APP:       "app",
-	PACKAGE_TYPE_BSP:       "bsp",
-	PACKAGE_TYPE_SDK:       "sdk",
 	PACKAGE_TYPE_COMPILER:  "compiler",
-	PACKAGE_TYPE_LIB:       "lib",
-	PACKAGE_TYPE_TARGET:    "target",
-	PACKAGE_TYPE_UNITTEST:  "unittest",
-	PACKAGE_TYPE_GENERATED: "generated",
 	PACKAGE_TYPE_MFG:       "mfg",
+	PACKAGE_TYPE_SDK:       "sdk",
+	PACKAGE_TYPE_GENERATED: "generated",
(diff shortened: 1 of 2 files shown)


utzig:
You need to fix indentation (use tabs instead of spaces!). Use `gofmt`.

ccollins476ad:
Hi Wanda,

Thanks for doing this.  It looks great.  Fabio is right, though - could you please run "gofmt -w" on:
* pkg/localpackage.go
* sysinit/sysinit.go

Go doesn't enforce whitespace layout, but it "strongly encourages" a particular style.  The gofmt tool reformats a go file so that it fits the recommended style.  I would never remember to do this, but my editor (vim) can be configured to run gofmt whenever you save a go file.

Thanks,
Chris

ccollins476ad:
Also, one really minor comment - try to restrict the first line in your git commit messages to 50 characters.  Git's short version of the log only prints the first 50 characters of each commit message.  It is best if you can briefly summarize the commit on this line.  Again, I wouldn't remember to do this, but my editor indicates when the first line is too long.

cwanda:
Hi Chris,  
Thanks for reviewing this! I ran gofmt  on the two files and reduced the length of first line of the commit message. Since the changes are so minor,  I just did a commit --amend instead of another commit.  The fix should be ready to go.  Thanks for all the go and git pointers.

ccollins476ad:
I'm sorry to be pedantic, but I still see some gofmt discrepencies.  I wonder if there are different versions of gofmt or something?  After running gofmt on the merged code, I have these diffs:

```
diff --git a/newt/pkg/localpackage.go b/newt/pkg/localpackage.go
index b34bd59..21fc43b 100644
--- a/newt/pkg/localpackage.go
+++ b/newt/pkg/localpackage.go
@@ -359,7 +359,7 @@ func (pkg *LocalPackage) Load() error {
                stage, err := strconv.ParseInt(stageStr, 10, 64)
                if err != nil {
                        return util.NewNewtError(fmt.Sprintf("Parsing pkg %s config: %s",
-                                                            pkg.FullName(), err.Error()))
+                               pkg.FullName(), err.Error()))
                }
                pkg.init[name] = int(stage)
        }
diff --git a/newt/sysinit/sysinit.go b/newt/sysinit/sysinit.go
index ddcc68b..29dc082 100644
--- a/newt/sysinit/sysinit.go
+++ b/newt/sysinit/sysinit.go
@@ -48,8 +48,8 @@ func buildStageMap(pkgs []*pkg.LocalPackage) map[int][]*initFunc {
                for name, stage := range p.Init() {
                        initFunc := &initFunc{
                                stage: stage,
-                               name: name,
-                               pkg: p,
+                               name:  name,
+                               pkg:   p,
                        }
                        sm[stage] = append(sm[stage], initFunc)
                }
@@ -63,7 +63,7 @@ func writePrototypes(pkgs []*pkg.LocalPackage, w io.Writer) {
        for _, p := range sorted {
                init := p.Init()
                for name, _ := range init {
-                       fmt.Fprintf(w, "void %s(void);\n", name);
+                       fmt.Fprintf(w, "void %s(void);\n", name)
                }
        }
 }
```

(i.e., gofmt normalized some whitespace and removed an extraneous semicolon.)

Does your version of gofmt leave these files unchanged?

Thanks,
Chris

ccollins476ad:
This is my mistake.  You didn't even change these two files.  Sorry about that.  I will merge this pull request.  Thanks again.

Chris

PR #34, opened 2017-01-28
Fixes MYNEWT-575: Checks for required "value" field in a setting definition. 

cwanda:
Updated newt to check that a setting definition has the required "value" field:
1) Checks the "value" key is specified for a setting definition map.
2) Prints an error message and aborts when one is not specified.

### DIFF
--- a/newt/syscfg/syscfg.go
+++ b/newt/syscfg/syscfg.go
@@ -283,9 +283,18 @@ func readSetting(name string, lpkg *pkg.LocalPackage,
 	entry := CfgEntry{}
 
 	entry.Name = name
-	entry.Description = stringValue(vals["description"])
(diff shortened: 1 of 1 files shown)


ccollins476ad:
Hi Wanda,

Functionally, it looks good to me.  I have a small comment, though: Go tends to use camelCase rather than underscores.  Would you mind changing the `value_val` and `value_exist` identifiers?

Thanks,
Chris

cwanda:
Hi Chris,
Thanks for looking at this and for the additional tip. I've committed the change to rename the two identifiers using camelCase convention.
Thanks,
Wanda

ccollins476ad:
Thanks, Wanda.

PR #35, opened 2017-02-04
MYNEWT-557 Warning override of undefined settings

cwanda:
Added support to newt build command to print out warning message for override of undefined settings.

### DIFF
--- a/newt/builder/targetbuild.go
+++ b/newt/builder/targetbuild.go
@@ -188,6 +188,13 @@ func (t *TargetBuilder) validateAndWriteCfg() error {
 		return util.NewNewtError(errText)
 	}
 
+	warningText := strings.TrimSpace(t.res.WarningText())
+	if warningText != "" {
+		for _, line := range strings.Split(warningText, "\n") {
+			log.Warn(line)
+		}
+	}
+
 	if err := syscfg.EnsureWritten(t.res.Cfg,
 		GeneratedIncludeDir(t.target.Name())); err != nil {
 
--- a/newt/builder/targetbuild.go
+++ b/newt/builder/targetbuild.go
@@ -191,7 +191,7 @@ func (t *TargetBuilder) validateAndWriteCfg() error {
 	warningText := strings.TrimSpace(t.res.WarningText())
 	if warningText != "" {
 		for _, line := range strings.Split(warningText, "\n") {
-			log.Warn(line)
+			log.Debugf(line)
 		}
 	}
 


ccollins476ad:
Hi Wanda,

I think this is a very useful change that will save developers from hours of debugging.  However, this change exposes an annoying issue with the way we've done syscfg.  A BSP defines its flash map in its bsp.yml file.  For this reason, it makes sense for the BSP package to specify what each flash area gets used for.  This is done by overriding settings defined by library packages.  For example, the nrf52dk package in @apache-mynewt-core overrides the following settings:
```
syscfg.vals:
    CONFIG_FCB_FLASH_AREA: FLASH_AREA_NFFS
    REBOOT_LOG_FLASH_AREA: FLASH_AREA_REBOOT_LOG
    NFFS_FLASH_AREA: FLASH_AREA_NFFS
    COREDUMP_FLASH_AREA: FLASH_AREA_IMAGE_1
```

The problem is that the BSP package doesn't know which library packages the app actually uses.  The BSP package overrides these settings just in case they are defined.  For example, if the app is built without core dump support, then the `COREDUMP_FLASH_AREA` setting will be undefined, which would elicit a warning on every build.

I'm not sure what the right fix is, but I'd be interested in hearing any thought you might have.  My current thinking is that we should adopt the following strategy:

1. For now, log these warnings to the debug log rather than the warning log.  That way, the user can still see the warnings if they use the `-ldebug` command line switch when they build.

2. Add some sort of `if-defined` directive to the syscfg syntax.  If an override specifies this directive, newt would ignore it when the specified setting is not defined.  We could add this directive to the flash area overrides in the BSP packages.  I think this is a rare case; these BSP settings are the only ones I can think of at the moment.

cwanda:
Hi Chris,
I see your point. I did see the warning for NFFS_FLASH_AREA when I was testing the code change.  I didn't look at it closely and thought it was a remnant from when I was creating errors to get error messages for the documentation.

1) I agree with your assessment. I think putting it in the debug log would be fine and I can make the change.  I could also add  a note about it in  "System Configuration and Initialization" Guide, if you think that would be helpful.

2) The if-defined directive is a nice idea if more packages will have a need for it. If this is isolated to the BSP packages only, then I think using the debug log is sufficient.

cwanda:
Hi Chris, 
I've updated the code to log the warning to the debug log instead now.  I will add a note about this to the system configuration and initialization validation and error messages section.  
What are your thoughts on whether we need the if-defined directive. I think it is probably sufficient to log the warning but am happy to add the support if you think it will be useful.
Thanks,
Wanda

sterlinghughes:
Hey -- do we have a resolution on this PR?

ccollins476ad:
Thanks, Wanda.

PR #36, opened 2017-02-12
add the pkg remove command to newt

sterlinghughes:


newt/cli/pkg_cmds.go
```diff
@@ -149,12 +150,64 @@ func pkgMoveCmd(cmd *cobra.Command, args []string) {
...
+	interfaces.SetProject(proj)
+
+	wd, err := os.Getwd()
+	if err != nil {
+		NewtUsage(cmd, util.ChildNewtError(err))
+	}
+
+	if err := os.Chdir(proj.Path() + "/"); err != nil {
+		NewtUsage(cmd, util.ChildNewtError(err))
+	}
```
utzig: You can do `defer os.Chdir(wd)` here and remove all other Chdir calls below.

PR #38, opened 2017-02-21
MYNEWT-543 Keep settings not set in syscfg variable

cwanda:
The current `newt target set target-name syscfg` variable saves the settings in a new target `syscfg.yml` file. Settings from the current target `syscfg.yml` that are not specified in the `syscfg`  variable are deleted. While setting a variable with only the values specified in the command is consistent with how other variables are set, the `syscfg` variable is a bit different because of multiple configuration settings.  A user might want to only override some settings but keep the values for the other settings in the target `syscfg.yml`.

This commit adds support to keep the values of the settings that are not specified in the set syscfg variable and to add or override the values in the target `syscfg.yml` for the settings specified for the variable. The code change converts the setting/value pair from the current syscfg.yml in to a KV String, and appends the KV String from command to it before it processes and saves the setting values. This way, an old setting value is either saved, or overridden if it is specified in the  `syscfg` variable.

A user who only wants to create a new target `syscfg.yml` with the settings specified in the `syscfg` variable can set  `syscfg=""` before it sets `syscfg` with the setting names and values. Example:
```
newt target set target_name syscfg="" syscfg=settingname1=settingvalue1:settingname2=settingvalue2 ...
```

### DIFF
--- a/newt/cli/target_cmds.go
+++ b/newt/cli/target_cmds.go
@@ -36,7 +36,7 @@ import (
 	"mynewt.apache.org/newt/newt/syscfg"
 	"mynewt.apache.org/newt/newt/target"
 	"mynewt.apache.org/newt/util"
-	"mynewt.apache.org/newt/viper"
+	//	"mynewt.apache.org/newt/viper"
 )
 
 var targetForce bool = false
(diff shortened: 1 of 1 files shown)


cwanda:
Chris, Aditi and I discussed this further and have decided to keep setting the `syscfg` variable behavior consistent with the other variables: setting a variable overwrites the current variable values.  We are looking into a feature enhancement, post release 1.0, that will support "appending" variable values.  

Closing this PR without code change.  The  `newt target set` documentation will be updated to let user know that the command will overwrite the current variable settings.

PR #39, opened 2017-02-22
newt: More detailed size report command

michal-narajowski:
This patch improves the output of the size command. The output is now similar to `make ram_report` and `make rom_report` in Zephyr. New flags were added for this purpose:

Flags:
  -F, --flash   Print FLASH statistics
  -R, --ram     Print RAM statistics

The size statistics are broken down into a tree-like structure, where
the leaves are symbols and branches are folders and files. For
each tree element there its size in bytes and percentage contribution
to the total size of the memory region.

```
Path                           Size       %
=============================================
(...)
libc                             32     0.49%
  baselibc                       32     0.49%
    src                          32     0.49%
      malloc.c                   32     0.49%
        __malloc_head            24     0.37%
        malloc_lock               4     0.06%
        malloc_unlock             4     0.06%
(...)
```

Moreover, there is more memory region info available after specifying
verbose flag(-v):

```
Mem FLASH: 0x00008000-0x00042000
Mem RAM:   0x20000000-0x20010000

Mem: FLASH
Name                       Size
.ARM.extab                    0
.ARM.exidx                   24
.imghdr                      32
.text                     18624
Total                     18680

Mem: RAM
Name                       Size
.bss                       5656
.stack_dummy                432
.vector_relocation          216
.data                       256
.bssnz                        0
Total                      6560
```

### DIFF
--- a/newt/builder/size.go
+++ b/newt/builder/size.go
@@ -106,8 +106,8 @@ func MakePkgSize(name string, memSections map[string]*MemSection) *PkgSize {
 /*
  * Go through GCC generated mapfile, and collect info about symbol sizes
  */
-func ParseMapFileSizes(fileName string) (map[string]*PkgSize, map[string]*MemSection,
-	error) {
+func ParseMapFileSizes(fileName string) (map[string]*PkgSize,
+	map[string]*MemSection, error) {
 	var state int = 0
 
 	file, err := os.Open(fileName)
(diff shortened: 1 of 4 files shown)


sterlinghughes:
Hi Michal, 

This is awesome -- thank you!  

The -l option to nm is not portable to Mac OS X.  I'm not sure what the OS X equivalent is, but as-is this doesn't run on Mac OS X.  When I remove -l it works, but doesn't show the path listings.  

Can you investigate if there is a cross platform option?  If not, can you make the -l option only run on Linux systems?

Thank You!

michal-narajowski:
Hi Sterling,

Thanks for the comment. Right now we don't have access to MacOS machine. I will make the -l option run only on Linux for now. I will try to fix it when I have access to MacOS.

michal-narajowski:
Hi Sterling,

I updated the commit.

Thanks,
Michał

michal-narajowski:
Hi Sterling,

I managed to get access to OSX machine and found out that we can use `arm-none-eabi-nm` instead of `nm` to make this feature compatible with both platforms. I fixed that and updated the pull request. 

Thanks,
Michał

utzig:
Just replying to the last comment without looking at whole thread so please forgive me...

`arm-none-eabi-nm` is only available to those who have gcc arm embedded installed. What about someone who is working a port for another architecture? Also why not just calling `nm` if it is already part of the gcc installation that the simulator uses?

utzig:
Ok, it's not from gcc. I probably have it from having Xcode tools installed. Also `objdump -t` provides similar info with different formatting.

utzig:
Also `brew install binutils` will also provide nm (named gnm in this case). Not sure what would be the best way to fix this so I'm ok with merging it this way and fixing later. Everybody right now probably has `arm-none-eabi`.

mkiiskila:
Should probably add nm to compiler.yml settings, so this can work for MIPS too.

PR #40, opened 2017-02-25
MYNEWT-639 Add help text and check for pkg name for newt pkg new command

cwanda:
1) Update help text for `newt pkg new` to have package name input.
2) Error check that a package name provided for the command.
3) Changed `libs/mylib` to `sys/mylib` in `newt pkg` help text to reflect latest source structure.
   (this change is not related to MYNEWT-639)
4) Added <package-name> argument to "newt pkg remove" Usage help text


cwanda:
Ammended the commit to:
1) Add <package-name> argument to the `newt pkg remove`  command usage help text.
2) Use package-name instead of package-location for `newt pkg new` command.

PR #44, opened 2017-03-01
MYNEWT-583; allow use of specific tag/repository revision in project.…

mkiiskila:
…yml.

This is done by using keyword 'tag' instead of 'vers' when specifying
repository details.

### DIFF
--- a/newt/project/project.go
+++ b/newt/project/project.go
@@ -181,15 +181,22 @@ func (proj *Project) upgradeCheck(r *repo.Repo, vers *repo.Version,
 		return false, err
 	}
 
-	branch, newVers, _ := rdesc.Match(r)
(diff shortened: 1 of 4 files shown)


mkiiskila:
newt install/newt upgrade is what I toyed with. I did not try anything more complicated. Do you have sample repositories set up for testing?

sterlinghughes:
No sample repos, but I think you'd want to make sure that you can recursively resolve dependencies  and detect versioning conflicts.  the sample repos i have setup, were setup a long time ago.  we should probably create a GH account that we run tests through regularly.

mkiiskila:
This needs a rework. This was starting to turn into spaghetti when dealing with dependencies. Closing this pr.

PR #48, opened 2017-03-03
MYNEWT-495: updates to sync command

utzig:
MYNEWT-495: updates to sync command

- When repo is already cloned, try to update from remote
- If "-f" is passed, create a modifications diff, and try harder to update current local (stash, clean, checkout, pull)

### DIFF
--- a/newt/cli/project_cmds.go
+++ b/newt/cli/project_cmds.go
@@ -20,6 +20,7 @@
 package cli
 
 import (
+	"fmt"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


utzig:
Kinda WIP because I'm still doing corner case testing...

utzig:
Just copy/paste of the last commit message:

```
- sync now creates a new branch when checking out a tag commit
- sync flow now:
  1) stash changes
  2) fetch branchs/tags from origin
  3) merges in changes on master and develop
  4) checks out old branch or new branch (creating if necessary)
  5) stash pops if the previous stash was succesful
- upgrade/install now try first to cleanup and existing repo and if it
  fails do the old style remove+clone.
```

Looks good on my tests, no issues so far. Seems good to be merged if no review issues are found.

sterlinghughes:
I don't think we should ever do the old style remove + clone, but I agree with the design otherwise.

utzig:
Well, hopefully the user will not pass `-f` all the time!

sterlinghughes:
I saw that after my comment, i think this commit is gtg.

utzig:
Btw, if I remember correctly `upgrade` always uses `force` == `true` which might be reworked in the future.

PR #52, opened 2017-03-07
Warns on unmatched pkg.name and pkg.path

utzig:
This adds a warning (when running in verbose mode) if a pkg.name is not equal to the current pkg path.


utzig:
Update to print the correct depender in the dependency fail message. Eg:

`Error: Could not resolve package dependency: @apache-mynewt-core/util/fakepkg; depender: apps/slinky`

PR #54, opened 2017-03-13
Make verbose compile more verbose...

utzig:
This changes compile do print the proper command output when running in verbose mode (-v).

Eg: `arm-none-eabi-gcc -D... bla.o ...`

Also removed a "duplicated" (kinda) link message.


utzig:
I haven't seen before the `ShellCommand` already prints the command if loglevel==debug. Will close this one for now.

PR #55, opened 2017-03-24
MYNEWT-583; newt - allow use of specific tag/repository revision.

mkiiskila:


newt/repo/repo.go
```diff
@@ -157,6 +173,32 @@ func NewRepoDependency(rname string, verstr string) (*RepoDependency, error) {
...
 
+func pickVersion(repo *Repo, versions []*Version) ([]*Version, error) {
+	fmt.Printf("Dependency list for %s contains a specific commit tag, " +
+		"so normal version number/stability comparison cannot be done.\n",
+		repo.Name())
+	fmt.Printf("If the following list does not contain the requirement to use, "+
+		"then modify your project.yml so that it does.\n")
+	for {
+		for i, vers := range versions {
+			fmt.Printf(" %d) %s\n", i, vers)
```
ccollins476ad: Should "vers" be converted to a string here?

mkiiskila: I believe it calls the String() method for it automagically, presumably due to %s on the format string.
However, I can add the explicit call to make it more clear what's going on.

mkiiskila:
Updated pull request.

ccollins476ad:
Thanks, Marko.  I don't think String() gets called automatically.  At least, it didn't when I tried it in the Go playground.  I did that test a while ago and no longer have the exact code I tested, though.

ccollins476ad:
Just an FYI- here is a playground program which I think reproduces this case: https://play.golang.org/p/UP8i8vo1oZ

```
package main

import (
	"fmt"
)

type S struct {
	a int
}

func (s *S) String() string {
	return fmt.Sprintf("%d", s.a)
}

func main() {
	s := S{5}
	fmt.Printf("s=%s\n", s)
}
```

Output:
```
s={%!s(int=5)}

Program exited.
```

Maybe there is something different in this example?

ccollins476ad:
Btw, I figured out why my playground example wasn't working.  The String() function operated on a `*S`, but I was passing a plain `S` to Printf.  I updated the example code and now it does work (https://play.golang.org/p/4kMNJ8NdEB):

```
package main

import (
	"fmt"
)

type S struct {
	a int
}

func (s *S) String() string {
	return fmt.Sprintf("%d", s.a)
}

func main() {
	s := &S{5}
	fmt.Printf("s=%s\n", s)
}
```

```
s=5

Program exited.
```

PR #58, opened 2017-05-01
Support for Windows Path to fix newt test problems

cwanda:
-  Convert paths to slashes for paths returned from `filepath.Dir` and `filepath.Glob` to make `newt test` work correctly.


aditihilbert:
Works good. Thanks!

PR #60, opened 2017-05-16
image erase command

jacobrosenthal:
Pairs with https://github.com/apache/incubator-mynewt-core/pull/281

Currently untested as Im getting the Killed: 9 go bug again, presumably after osx 10.12.5 update


jacobrosenthal:
OK tested. Had to update to go 1.8.1. This works for me, though note Im I never get the rc response as my nrf51 device disconnects. (as expected) during the erase.

PR #61, opened 2017-05-22
[WIP]MYNEWT-640 Add support for newt target append command.

cwanda:
The newt target append command allows the user to append target variable values
for variables that can take multiple values and not overwrite other variable values. This addresses the concern raised in MYNEWT-543.
The append command uses the same format as `newt target set`.
The variables that can have values appended are: syscfg, cflags, aflags, and lfags.

Note: We need to decide how to handle deleting a value. Currently, for the
`newt target set` command, when the user enters `<variable-name>=""`, the
variable-name is deleted from the yml file.   The append command
implementation flags an error when the user specifies
`<variable-name>=""`  (since we don't want to overload the command),
and leaves it to the user to use the set command to delete the variable.
However, there currently isn't a way for the user to delete a
specific value for the multi-value. Couple of ways we can support delete:
1) Add a newt target delete command for deleting variables:
   - If the user enters `<variable-name>=""`, then the variable is deleted
      from the yml file.
   - For multi-value variables, the user can specify the value to delete as
     follows:
      *  `syscfg=LOG_LEVEL:CONFIG_NEWTMGR` deletes the LOG_LEVEL and
         CONFIG_NEWTMGR variable from the target syscfg.yml
      * `cflags="-DNODEBUG -DMYDEFINE"` deletes the -DNODEBUG and -DMYDEFINE
        flags.
   - Remove the variable deletion functionality from newt target set.

2) Don't support deleting individual values  for multi-values variable. User
would use the newt target set command to delete the variable and then enter the new values.

3) Other options such as overloading the append command somehow?

I think option 1 would be the cleanest solution. But not sure if we wan't so many newt target sub commands.

### DIFF
--- a/newt/cli/target_cmds.go
+++ b/newt/cli/target_cmds.go
@@ -42,6 +42,9 @@ import (
 
 var targetForce bool = false
 
+// target variables that can have values appended with the the append command.
+var appendVars = []string{"aflags", "cflags", "lflags", "syscfg"}
+
 func resolveExistingTargetArg(arg string) (*target.Target, error) {
 	t := ResolveTarget(arg)
 	if t == nil {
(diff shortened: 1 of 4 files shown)


cwanda:
Will change command to amend and add a -d flag to delete variables.

cwanda:
Closing the PR. Not sure why commit https://github.com/apache/incubator-mynewt-newt/pull/61/commits/7408d18499855f6ba253d7fe90d1736e0bf85711 which has already been merged in  a different is included in this pr

PR #65, opened 2017-05-26
MYNEWT-649 Validate target variable target set command.

cwanda:
Valid variables are: aflags, app, build_profile, bsp, cflags, lflags, loader, syscfg

### DIFF
--- a/newt/cli/target_cmds.go
+++ b/newt/cli/target_cmds.go
@@ -860,7 +860,7 @@ func AddTargetCommands(cmd *cobra.Command) {
 	setHelpEx += "syscfg=LOG_NEWTMGR=1:CONFIG_NEWTMGR=0\n"
 
 	setCmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


newt/cli/target_cmds.go
```diff
@@ -684,7 +704,7 @@ func AddTargetCommands(cmd *cobra.Command) {
 
 	setCmd := &cobra.Command{
 		Use: "set <target-name> <var-name>=<value>" +
-			"[:<var-name>=<value>...]",
+			"[<var-name>=<value>...]",
```
ccollins476ad: Good catch.  However, there is still a minor issue here.  To be pedantically correct, there should be a space after the first '>' (i.e., before '[').

cwanda:
Hi Chris,
Good catch. I added the space.  Thanks  for your reviews of all the newt prs.

PR #66, opened 2017-06-07
image: Add support for RSA-PSS signatures

d3zd3z:
Currently, RSA signatures are done with PKCS#1 v1.5.  Newer versions of
this standard (starting with 2.1) define a newer signing algorithm, PSS,
and enourage adoption of this algorithm over the older method.

Add support for this to the 'newt' tool so that create-image can sign
RSA with PSS.  Since the keys are the same, add a command line argument
`--rsa-pss` to use the new algorithm.  Once this is adopted more fully
by the bootloader this could be made the default.

### DIFF
--- a/newt/image/image.go
+++ b/newt/image/image.go
@@ -46,6 +46,10 @@ import (
 	"mynewt.apache.org/newt/util"
 )
 
+// Set this to enable RSA-PSS for RSA signatures, instead of PKCS#1
+// v1.5.  Eventually, this should be the default.
+var UseRsaPss = false
+
 type ImageVersion struct {
 	Major    uint8
 	Minor    uint8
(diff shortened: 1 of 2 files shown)


newt/image/image.go
```diff
@@ -46,6 +46,10 @@ import (
 	"mynewt.apache.org/newt/util"
 )
 
+// Set this to enable RSA-PSS for RSA signatures, instead of PKCS#1
+// v1.5.  Eventually, this should be the default.
+var UseRsaPss = false
```
mkiiskila: This would look better if this setting was a property of Image, as that's where all the other data associated with the signing process is being collected.

d3zd3z: I'm not quite sure how this would work. This is the point in the process where we are creating the image header, so none is available yet. Other arguments specify the version, for example.

Other than this option, the signing algorithm comes from the key that the user specified. However with v1.5 vs pss, the same keys are used, and we need some way to specify how this key should be used.

mkiiskila: Ah, I forgot that there was a layer of indirection added between the CLI and image signing for split images. Reaching through that will be a bit cumbersome.

PR #72, opened 2017-06-22
MYNEWT-553; add option to print only executed commands during build

michal-narajowski:
To print only executed commands:
newt build --silent --printCmds <target>

### DIFF
--- a/newt/vendor/mynewt.apache.org/newt/util/util.go
+++ b/newt/vendor/mynewt.apache.org/newt/util/util.go
@@ -42,6 +42,7 @@ import (
 )
 
 var Verbosity int
+var PrintShellCmds bool
 var logFile *os.File
 
 func ParseEqualsPair(v string) (string, string, error) {
@@ -247,6 +248,7 @@ func Init(logLevel log.Level, logFile string, verbosity int) error {
 	}
 
 	Verbosity = verbosity
+	PrintShellCmds = false
 
 	return nil
 }
@@ -290,6 +292,10 @@ func ShellCommandLimitDbgOutput(
 		envLogStr = strings.Join(env, " ") + " "
 	}
 	log.Debugf("%s%s", envLogStr, strings.Join(cmdStrs, " "))
+
+	if PrintShellCmds {
+		StatusMessage(VERBOSITY_SILENT, "%s\n", strings.Join(cmdStrs, " "))
+	}
 
 	name := cmdStrs[0]
 	args := cmdStrs[1:]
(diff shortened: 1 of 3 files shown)


newt/vendor/mynewt.apache.org/newt/util/util.go
```diff
@@ -42,6 +42,7 @@ import (
 )
 
 var Verbosity int
+var PrintBuildCmds bool
```
ccollins476ad: Since the util library is not just used by newt (e.g., it is used by newtmgr), I would rename this variable to something more generic such as `PrintShellCmds`.

Also, the changes look good to me, but the changes to util.go were only made to the vendored version of this file.  The Go vendoring system is kind of messy, and it creates a lot of hassle when all you want to do is make a small change.

Basically, there are several versions of util.go:
* util/util.go
* newt/vendor/mynewt.apache.org/newt/util/util.go
* newtmgr/vendor/mynewt.apache.org/newt/util/util.go

The first one (util/util.go) is the "master".  The others are just copies that get distributed with the tool that depends on it.

The correct way to update util.go is to make your changes to util/util.go, and then *revendor* the tools which require the updated util package.  This involves running the `godep` tool, and can be a bit tricky to do when everything is in the same repo, as in this case.  It is important to make sure your changes go into util/util.go, or else they will be lost the next time someone revendors newt.

michal-narajowski: Could you help me with the revendor? I cannot figure out how to do it.

ccollins476ad: Hmm... the godep tool seems to be pretty much broken, at least for our use case.  There are two options that will work:

1. Perform a complete revendoring of newt.
2. Just copy the util.go changes by hand.

I think option 2 is preferred here, since we don't want to upgrade all the dependencies when the current ones are working.  The problem is that the Godeps.json file will no longer be correct, but since the tool doesn't work for us anyway, I don't see much value in this file!  So, I would just copy your updated util.go file to util/util.go and commit it.

For the record, here is how you would do a full revendoring:
1. `rm -rf Godeps vendor`
2. `git commit -a -m 'revendor newt'`
3. `godep save`
4. `git add Godeps vendor`
5. `git commit -a -m 'revendor newt'`
6. Use `git rebase -i` to fixup the two revendor commits

I think we need to look into a tool other than godep to do our vendoring.  It looks like godep hasn't been updated for a while.

michal-narajowski: Thanks, Chris. I updated the commit by copying util.go to util/util.go.

ccollins476ad: Cool, looks good.  What do you think about renaming `PrintBuildCmds` to `PrintShellCmds` (or similar)?  Other tools use the util library, but they don't build anything.

michal-narajowski: Sure, updated.

PR #74, opened 2017-06-22
syscfg - Allow a package to override its own setting

ccollins476ad:
(Jira: https://issues.apache.org/jira/browse/MYNEWT-790)

syscfg - Allow a package to override its own setting.

PKG-A can only override a setting defined by PKG-B if PKG-A has a
greater priority than PKG-B.  If PKG-A's priority is less than or equal
to PKG-B's, the override attempt is rejected and reported as a priority
violation.

This commit relaxes that rule slightly: a package can override settings
that it itself defines.

This behavior is needed  for making a setting's default value
conditional on another setting.  For example:

```
    syscfg.defs:
        TIMER_0:
            description: 'NRF52 Timer 0'
            value:  1
        TIMER_3:
            description: 'NRF52 Timer 3'
            value:  0

    # Use timer 3 instead of timer 0 if target wants to use the low
    # power clock.
    syscfg.vals.BLE_LP_CLOCK:
        TIMER_0: 0
        TIMER_3: 1
```

### DIFF
--- a/newt/resolve/resolve.go
+++ b/newt/resolve/resolve.go
@@ -362,16 +362,16 @@ func (r *Resolver) reloadCfg() (bool, error) {
 		return false, err
 	}
 
-	changed := false
+	// Determine if any settings have changed.
 	for k, v := range cfg.Settings {
 		oldval, ok := r.cfg.Settings[k]
 		if !ok || len(oldval.History) != len(v.History) {
 			r.cfg = cfg
-			changed = true
+			return true, nil
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rymanluk:
I works for me.
In controller/syscfg.yml I have

syscfg.defs:
    BLE_LL_CFG_FEAT_LL_EXT_ADV:
        value: 0

syscfg.vals.BLE_EXT_ADV:
    BLE_LL_CFG_FEAT_LL_EXT_ADV: 1

PR #77, opened 2017-07-11
Create compile_commands.json file on build

michal-narajowski:
Can be useful for clang tools, rtags, etc.

For example:
To perform static analysis on a source code:

```
clang-check -analyze -p path/to/compile_commands.json path/to/source/file
```

or

```
find path/to/sources/ -name '*.c'|xargs clang-check -analyze -p path/to/compile_commands.json
```

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -20,13 +20,16 @@
 package builder
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
 
 	"mynewt.apache.org/newt/newt/image"
+	"mynewt.apache.org/newt/newt/interfaces"
 	"mynewt.apache.org/newt/newt/newtutil"
 	"mynewt.apache.org/newt/newt/pkg"
 	"mynewt.apache.org/newt/newt/repo"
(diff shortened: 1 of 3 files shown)


newt/builder/build.go
```diff
@@ -568,6 +571,35 @@ func (b *Builder) Build() error {
...
+	}
+
+	cmdBytes, err := json.MarshalIndent(compileCommands, "", "    ")
+	if err != nil {
+		log.Error("Unable to encode compilation commands as JSON")
+		return nil
+	}
+
+	outDir := filepath.Dir(b.AppElfPath())
+	cmdPath := outDir + "/compile_commands.json"
```
ccollins476ad: It might be better to define a function in `builder/paths.go` which returns the compile commands json path (CompileCmdsPath() or similar).  That way, all the paths are defined in the same place.

newt/builder/build.go
```diff
@@ -568,6 +571,35 @@ func (b *Builder) Build() error {
...
+	errWrite := ioutil.WriteFile(cmdPath, cmdBytes, 0644)
+	if errWrite != nil {
+		log.Error("Unable to write compile_commands.json file")
+		return nil
```
ccollins476ad: I think this failure should abort the build (i.e., return a NewtError here).  Also, I would include the original error text in the message so that the user knows why file creation failed.

michal-narajowski:
Thanks, Chris. Updated.

michal-narajowski:
Hi, @ccollins476ad.
Should I merge the PR or do you want to review the new changes?
BR

PR #84, opened 2017-08-08
Add option to launch build command using /bin/sh

andrzej-kaczmarek:
This adds "--executeShell" option to "newt build" which will run build
commands using "/bin/sh -c" instead of directly running the compiler.

This is useful e.g. for Coverity which does not recognize compiler
properly when launched from newt (could be due to clone() instead of
fork(), but not really sure). When running using shell, compiler is
properly forked and recognized.

This is applicable only for Linux and MacOS platforms.


sjanc:
could we also have this switch in "newt test" (when it builds tests)?

PR #88, opened 2017-08-14
[RFC] Add option to generate CMakeLists.txt file for a target

michal-narajowski:
This PR adds a command to generate CMakeLists.txt file for a target. The file is placed in project root directory. 

Usage: newt target cmake <target>

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -447,8 +447,8 @@ func (b *Builder) PrepBuild() error {
 	baseCi.AddCompilerInfo(bspCi)
 
 	// All packages have access to the generated code header directory.
-	baseCi.Cflags = append(baseCi.Cflags,
-		"-I"+GeneratedIncludeDir(b.targetPkg.rpkg.Lpkg.Name()))
+	baseCi.Includes = append(baseCi.Includes,
+		GeneratedIncludeDir(b.targetPkg.rpkg.Lpkg.Name()))
 
 	// Note: Compiler flags get added at the end, after the flags for library
 	// package being built are calculated.
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -311,6 +311,11 @@ func (b *Builder) collectCompileEntriesBpkg(bpkg *BuildPackage) (
 	return entries, nil
 }
 
+func (b *Builder) CollectCompileEntriesBpkg(bpkg *BuildPackage) (
+	[]toolchain.CompilerJob, error) {
+	return b.collectCompileEntriesBpkg(bpkg)
+}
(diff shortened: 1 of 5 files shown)


ccollins476ad:
Overall, it looks good to me.  However, I have a few comments.  A lot of these are pretty minor, so you can use your own discretion when deciding whether to make any changes.

1. Writing `CMakeLists.txt` to the project directory concerns me just a little bit, for a few reasons:
a. It "pollutes" the project directory with a generated file.  Aside from this file, newt writes all its files to the `bin` directory.
b. It limits the user to one target at a time.  Other commands, e.g., `build`, allow multiple target names to be specified for a single invocation.  When automating builds, it is useful to be able to specify multiple targets or `all`.

I am not sure how people would end up using cmake with newt, so the above concern may be invalid.  If this feature is intended for manual use only, then the current implementation might be perfect.

2. The help text for `target cmake` looks a bit off: `Cmake all the variables for the target specified by <target-name>.`

3. The `CMakeTargetWrite()` and `CMakeBuildPackageWrite()` functions should use `StatusMessage()` to display text rather than `fmt.Printf()`.  That way, the user can silence the output if desired.  I would use a verbosity of `util.VERBOSITY_DEFAULT` here.

4. The above two functions print `Generating code for <package>`.  That seems slightly inaccurate since no code is generated.  I'm afraid I don't have any suggestions here :).

5. `cmake.go` redefines the compiler types.  Would it be better to just use the original definitions in the `toolchain` package (e.g., `toolchain.COMPILER_TYPE_C`)?

6. In `CMakeTargetGenerate()`:
```
    CmakeFileHandle, _ := os.Create(CmakeListsPath())
```

It seems dangerous to ignore the error here.  I suggest checking the error and returning `ChildNewtError(err)` on failure.  It is important to wrap the original error with `ChildNewtError()` in case the calling code needs to know the specific file system error that caused the problem.

7.  In `CMakeTargetGenerate()`:

```
    targetBuilder, err := NewTargetBuilder(target)
    if err != nil {
        return util.NewNewtError(err.Error())
    }
```

`NewTargetBuilder()` already returns a NewtError, so this code can just return `err` unadorned.  By returning the original error, you preserve the stack trace.

michal-narajowski:
I fixed issues 2-7. As for the 1. I agree that it pollutes the project directory. Initially, I wanted to generate a file for each target, but CLion searches for CMakeLists.txt in the root directory and it didn't work for me. I'm not sure if it is something you can change in CLion, so I decided to change the path to root path. I hope that we can tweak this as more people start using this feature and sharing their experience.

PR #91, opened 2017-08-30
Makefile for building debian package

cwanda:
So we don't have to create a Makefile each time we build a deb package for newt


cwanda:
Thanks chris. I also updated the Makefile with comments  for newtmgr

PR #94, opened 2017-09-05
New img format

utzig:
Ready for merge.


mkiiskila:
Regarding comment 2: this is a hash of the public key used in signing. The hash is computed using the same algorithm, the number of bytes being hashed depends on the size of the key (aka 224 vs 256).

utzig:
Removed the spurious comment in a new squashed commit.

PR #95, opened 2017-09-06
unixchild - fix race condition

ccollins476ad:
This commit gives the unixchild ErrChan a depth of 1 (was 0).  Prior to this change, there was a race condition when the child process terminated at the same time that the calling code stopped the unixchild library.

A depth of 1 is sufficient because there will never be more than one error.  When the first error is encountered, the library terminates.


spoonofpower:
👍

PR #96, opened 2017-09-11
newt: Workaound incorrectly resolved symbols in 'newt size'

andrzej-kaczmarek:
For unknown reason arm-none-eabi-nm does not resolve source files
properly for some symbols - it just leaves them without filename.
As a result 'newt size' put them in '(other)' group which is not
really helpful.

This patch adds workaround for this by calling 'addr2line' on symbols
with missing filename and this tool seems to have no problem with
finding proper filename.

### DIFF
--- a/newt/builder/size_report.go
+++ b/newt/builder/size_report.go
@@ -35,7 +35,7 @@ func runNmCommand(elfFilePath string) ([]byte, error) {
 		err    error
 	)
 	cmdName := "arm-none-eabi-nm"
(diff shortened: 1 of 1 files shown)


mkiiskila:
You might want to take a look at how this is done at newt/builder/size.go just by parsing the list file. That data output is placed in the manifest file.

utzig:
Is this waiting anything?

andrzej-kaczmarek:
since this works fine and fixes the problem I'll merge this now and will take another look (as suggested in comment) later

PR #103, opened 2017-10-26
Newt - Generic git repositories

ccollins476ad:
This commit adds support for a new repository type, as specified in the `project.yml` file: "git".  Here is an example:

```
        type: git
        vers: 0-latest
```

Repositories of this type do not have any fields for credentials; security is handled by the `git` tool, not newt.  The example repo above would authenticate via ssh.

In addition, this commit cleans up some inconsistent and unintuitive behavior in the following newt commands:

* `install`
* `upgrade`
* `sync`

The new behaviors of these commands are described below:

* `install`:
    * For each repo in `project.yml`:
        * If `-f` (force) specified, delete repo's directory.
        * If repo doesn't exist:
            * Clone repo.
            * Checkout appropriate branch.
            * Populate `project.state` file.

* `upgrade`:
    * For each repo in `project.yml`:
        * If repo doesn't exist:
            * Clone repo.
            * Checkout appropriate branch.
            * Populate `project.state` file.
        * Else:
            * If versions in `project.state` and `project.yml` don't match:
                * Prompt user [Y/n]
                * If any local changes:
                    * If !force:
                        * Print error and abort.
                * Upgrade repo to correct branch.  If local changes, attempt merge.
                * Populate `project.state` file.

* `sync`:
    * For each repo in `project.yml`:
        * If repo not installed:
            * Print error message.
        * Else:
            * If remote contains changes not in local branch:
                * If any local changes:
                    * Save `git diff` output to temp file.
                * Replace repo contents with latest from remote.

### DIFF
--- a/newt/downloader/downloader.go
+++ b/newt/downloader/downloader.go
@@ -21,9 +21,7 @@ package downloader
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -36,7 +34,7 @@ import (
 )
 
 type Downloader interface {
-	FetchFile(name string, dest string) error
+	FetchFile(path string, filename string, dstDir string) error
 	Branch() string
 	SetBranch(branch string)
 	DownloadRepo(branch string) (string, error)
@@ -44,10 +42,14 @@ type Downloader interface {
 	UpdateRepo(path string, branchName string) error
 	CleanupRepo(path string, branchName string) error
 	LocalDiff(path string) ([]byte, error)
+	AreChanges(path string) (bool, error)
 }
 
 type GenericDownloader struct {
 	branch string
+
+	// Whether 'origin' has been fetched during this run.
+	fetched bool
 }
 
 type GithubDownloader struct {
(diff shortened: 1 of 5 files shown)


newt/downloader/downloader.go
```diff
@@ -240,17 +278,19 @@ func (gd *GithubDownloader) authenticatedCommand(path string,
...
 		// Use the github API
 		url = fmt.Sprintf("https://%s/api/v3/repos/%s/%s/%s?ref=%s",
-			gd.Server, gd.User, gd.Repo, name, gd.Branch())
+			gd.Server, gd.User, gd.Repo, filename, gd.Branch())
 	} else {
 		// The public github API is ratelimited. Avoid rate limit issues with
 		// the raw endpoint.
 		url = fmt.Sprintf("https://raw.githubusercontent.com/%s/%s/%s/%s",
-			gd.User, gd.Repo, gd.Branch(), name)
+			gd.User, gd.Repo, gd.Branch(), filename)
```
spoonofpower: Is an http request still needed to fetch the file? With the other changes in this PR, isn't the file now already on disk? Can the GithubDownloader FetchFile be implemented in the same was as the GitDownloader FetchFile?

spoonofpower:
This worked great for me.

ccollins476ad:
Thanks for taking a look, @spoonofpower.  I wasn't able to split this into logically independent commits, so this probably was a pain to look at.

You are correct about the github API being unnecessary now.  That was a good catch.  I have changed `GithubDownloader#FetchFile` to look like `GitDownloader#FetchFile`.

PR #111, opened 2017-11-28
fixed matchNamePath for windows and added -f flag for rm in build.sh

cwanda:
fixed matchNamePath for windows  and also added -f to rm /tmp/* in build.sh so
 won't get prompt when removing write protected file.


cwanda:
hi fabio. Thanks.  Not sure if you are waiting to merge later, but in case you are not, you will need to merge the PR for me. I don't have rights to merge.

PR #113, opened 2017-11-28
Update release to 1.3.0

utzig:
Signed-off-by: Fabio Utzig <utzig@apache.org>


mkiiskila:
Should this be merged to 1_3_0_dev branch instead?

utzig:
Yes, but according to the document I need to update the blinky tag it seems. Is not this usually just pushed, instead of PRing?

PR #115, opened 2017-12-05
make master newt version ahead of releases

placrosse:
When doing a brew install from --HEAD, resulting newt should not show version older than existing releases and candidates?

### DIFF
--- a/newt/newtutil/newtutil.go
+++ b/newt/newtutil/newtutil.go
@@ -35,8 +35,8 @@ import (
 	"mynewt.apache.org/newt/viper"
 )
 
-var NewtVersion Version = Version{1, 2, 0}
-var NewtVersionStr string = "Apache Newt version: 1.2.0-dev"
+var NewtVersion Version = Version{1, 4, 0}
+var NewtVersionStr string = "Apache Newt version: 1.4.0-dev"
 var NewtBlinkyTag string = "master"
 var NewtNumJobs int
 var NewtForce bool


utzig:
I believe the process is to merge back the `x_x_x_dev` branch that originated the release candidate to `master` once the new release is accepted. This patch would add an extra step of always updating `master` after a new release branch is created and would cause a conflict when doing the merge. This cannot be just a code change, it would require an update to the release process.

placrosse:
I see your point.  I found it confusing to have newt version return 1.2.0-dev this morning, when 1.2.0 has been released for weeks.  I assumed that 1.2.0-dev comes prior to 1.2.0 (release), but it is actually after at present as you described, correct?

ccollins476ad:
I think the version number has been wrong in newt (perhaps we forgot to update it).  In my opinion, it is most logical if the dev version number is greater than the last release.  So, after 1.3 is released, I would expect newt from master to report a version number of `1.4.0-dev`.

mkiiskila:
+1, I agree

utzig:
+1

utzig:
@placrosse Please rebase! I agree with @ccollins476ad that it would be better to use `Version{1, 3, 9999}` to differentiate from the release version.

sterlinghughes:
have we ever merged this?  Can we either rebase ourselves and merge or close the PR?

utzig:
Superseded by #146.

PR #119, opened 2017-12-21
Fix pbkdf2 hash

utzig:
This needs to be rebased when #108 is merged!

This allows for pbkdf2 that are not using `sha1` for hashing.

/cc @d3zd3z

### DIFF
--- a/newt/image/encrypted.go
+++ b/newt/image/encrypted.go
@@ -24,20 +24,25 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/sha1"
+	"crypto/sha256"
 	"crypto/x509"
 	"crypto/x509/pkix"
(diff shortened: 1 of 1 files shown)


newt/image/encrypted.go
```diff
@@ -147,39 +147,9 @@ func unwrapPbes2Pbkdf2(param *pbkdf2Param, size int, iv []byte, hashNew hashFunc
 	plain := make([]byte, len(encrypted))
 	enc.CryptBlocks(plain, encrypted)
 
-	plain, err = checkPkcs7Padding(plain)
```
d3zd3z: It "works", but it then doesn't detect messages with corrupt padding. We really do need to verify that padding after performing the decryption.

utzig:
Updated to leave padding removal code untouched!

PR #131, opened 2018-02-02
MYNEWT-882 Newt - Allow spaces in paths

ccollins476ad:
Newt was interpreting each include path as a space-delimited list of directories.  If the project directory name contained a space, the path would get split into two separate paths, causing an `#include` directive to fail.

Now, newt treats each entry as a single include path.


ccollins476ad:
There is still an issue here.  Newt fails to parse the generated `.d` files that contain spaces.

PR #137, opened 2018-02-20
 Don't require repo TOC in project.yml

ccollins476ad:
The user's `project.yml` file needed to specify each repo dependency
twice:
* Name in `project.repositories` table of contents
* Definition as separate item (`repository.<...>`)

This made it a bit of a hassle to add a new repo to a project.

This commit eliminates the need for the `project.repositories` table of
contents.  Now newt assumes all top-level items starting with
"repository." describe a repository, and it attempts to load them.


utzig:
Good move!

PR #138, opened 2018-02-23
newt - Don't abort if setting is redefined.

ccollins476ad:
Syscfg settings must have unique names.  When newt detected two settings
with the same name, it would terminate with this error message:
```
setting <name> redefined
```

This commit changes two things:

1. The error is now just a warning.  Newt warns the user about the
issue, but ignores the second definition and proceeds with its
operation.

Allowing newt to proceed makes it easier to debug the syscfg issue.  Now
the user can use commands like `target revdep` to debug the issue.

2. The message text is now more detailed:
```
* Warning: setting <name> redefined (pkg1=<pkg-name> pkg2=<pkg-name>)
```


utzig:
LGTM

PR #140, opened 2018-03-06
newt - Repo dependencies

ccollins476ad:
Allow a repo to specify which other repos it requires in its `repository.yml` file.  For example, a `repository.yml` file might contain the following:

```
    repository.deps:
        mynewt-nimble:
            type: git
            url: 'git@github.com:apache/mynewt-nimble.git'
            vers:
                master:             0-dev
                mynewt_1_4_0_tag:   >=1.4.0

        mynewt-nffs:
            type: git
            url: 'git@github.com:apache/mynewt-nffs.git'
            vers:
                master:             0-dev
                mynewt_1_4_0_tag:   1.4.0
```

What this says is:
* The master branch of of this repo depends on:
    * nimble 0-dev
    * nffs 0-dev
* The mynewt_1_4_0_tag tag of of this repo depends on:
    * nimble 1.4.0 or later
    * nffs 1.4.0 only

When the user performs an install, upgrade, or sync operation, newt includes the depended-on repos in the operation.  For example, if the user tells newt to install version 1.4.0 of the above repo, newt will also install suitable versions of nimble and nffs.  If the user then upgrades the above repo to 0.0.0, both nimble and nffs will be upgraded to 0.0.0 along with it.

If the user's request cannot be completed due to conflicting repo version requirements, newt informs the user of the problem and leaves the project unchanged.

In addition, the `-a` (`--ask`) flag has been added to the install, upgrade, and sync commands.  When this flag is specified, newt describes the operation it is about to perform and prompts the user for confirmation before proceeding.

newt had an existing implementation of this feature that this PR replaces.  There were a number of things that the old implementation wasn't doing correctly (it may well have worked when it was originally added):

* Issues with repos whose dependency list changes with version.  Each repo would initially load a set of dependencies corresponding to the currently-installed version.  This could be problematic when a dependent repo is upgraded, as the depended-on repos might no longer satisfy the dependent's requirements.

* Failure to treat version aliases as equivalent.  For example, the `mynewt_arduino_zero` repo maps both 0.0.0 and 0.0.1 to the master branch, making both versions identical.  This could result in conflicts being reported when there actually were none.

Finally, sorry about the size of the "Repo dependencies" commit in this PR!  I tried to break it up into something that would be easier to review, but I found it impossible.


sterlinghughes:
@ccollins476ad -- do you have some sample repositories and projects that I can try this out on to test it?  The PR is pretty big :-) . I've looked through it, but honestly I'd like to try it out before approving and merging it.

ccollins476ad:
That is a good point.  I will add some test cases with results in a follow up comment to this PR.

ccollins476ad:
By the way, I noticed a few issues while running these tests.  I have force pushed the fixes.  The issues were:

* `newt install` might install repos that weren't actually required.
* A repo's `repository.yml` file might be fetched more than once.  This should only be done once per repo.
* Newt would prefer a repo's currently-installed version over a new one.  This is incorrect; `newt upgrade` should always upgrade to the latest non-conflicting version.

ccollins476ad:
Here are some tests with results.  I created some empty Mynewt repos for these tests ("mynewt-repo-01", "mynewt-repo-02").  These repos are in my github (ccollins476ad), but they should be publicly accessible to others.

## Test 1
#### `project.yml`
```
repository.mynewt-repo-01:
    type: git
    vers: 0.0.0
    url: 'git@github.com:ccollins476ad/mynewt-repo-01.git'
```

#### Action
```newt install```

#### Result
```
install mynewt-repo-01 (0.0.0)
install mynewt-repo-02 (0.0.0)
install apache-mynewt-core (0.0.0)
```

#### Notes
`project.yml` requires mynewt-repo-01,==0.0.0
mynewt-repo-01,0.0.0 requires mynewt-repo-02,==0-dev (0.0.0)
mynewt-repo-02,0.0.0 requires apache-mynewt-core,==0-dev (0.0.0)

## Test 2
`project.yml`
```
repository.mynewt-repo-01:
    type: git
    vers: 1.3.0
    url: 'git@github.com:ccollins476ad/mynewt-repo-01.git'
```

#### Action
```newt upgrade```

#### Result
```
upgrade to mynewt-repo-01 (0.0.0 --> 1.3.0)
upgrade to mynewt-repo-02 (0.0.0 --> 1.3.0)
install mynewt_arduino_zero (1.3.0)
```

#### Notes
`project.yml` requires mynewt-repo-01,==1.3.0
mynewt-repo-01,1.3.0 requires mynewt-repo-02,>=1.2.0
mynewt-repo-02,1.3.0 requires mynewt_arduino_zero,==1.3.0

## Test 3
`project.yml`
```
repository.mynewt-repo-01:
    type: git
    vers: 1.3.0
    url: 'git@github.com:ccollins476ad/mynewt-repo-01.git'
repository.mynewt_arduino_zero:
    type: git
    vers: 1.2.0
    url: 'git@github.com:runtimeco/mynewt_arduino_zero.git'
```

#### Action
```newt upgrade```

#### Result
```
Skipping "mynewt-repo-01": already upgraded (1.3.0)
upgrade mynewt-repo-02 (1.3.0 --> 1.2.0)
install mcuboot (1.1.0)
upgrade mynewt_arduino_zero (1.3.0 --> 1.2.0)
```

#### Notes
`project.yml` and mynewt-repo-02,1.3.0 require conflicting versions of mynewt_arduino_zero.  The `project.yml` requirements always take precedence over repo dependencies, so the issue is resolved by downgrading mynewt-repo-02 to a version that doesn't depend on a conflicting version of mynewt_arduino_zero.

## Test 4
`project.yml`
```
repository.mynewt-repo-01:
    type: git
    vers: 0.0.0
    url: 'git@github.com:ccollins476ad/mynewt-repo-01.git'
    type: git
    vers: 1.3.0
```

#### Action
```newt upgrade```

#### Result
```
Error: Repository conflicts:
    Installation of repo "apache-mynewt-core" is blocked:
                   `project.yml` requires apache-mynewt-core ==1.3.0
              mynewt-repo-02,0.0.0 requires apache-mynewt-core ==0.0.0
```

## Test 5
`project.yml`
```
repository.mynewt-repo-01:
    type: git
    vers: 1.3.0
    url: 'git@github.com:ccollins476ad/mynewt-repo-01.git'
    type: git
    vers: 1.3.0
```

#### Action
```newt upgrade```

#### Result
```
Skipping "mynewt-repo-01": already upgraded (1.3.0)
upgrade apache-mynewt-core (0.0.0 --> 1.3.0)
upgrade mynewt-repo-02 (1.2.0 --> 1.3.0)
upgrade mynewt_arduino_zero (1.2.0 --> 1.3.0)
```

#### Notes
Repos are always upgraded to the latest version possible.  Now that the mynewt_arduino_zero requirement has been removed from `project.yml`, mynewt-repo-02 is allowed to upgrade back to 1.3.0.

File: newt/repo/repo.go [hunk 1]
```diff
@@ -3,7 +3,7 @@
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
  * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
+ * to you under the Apache License, newtutil.RepoVersion 2.0 (the
```
utzig: `find/replace` gone wrong! ;-)

newt/project/project.go
```diff
@@ -198,177 +209,350 @@ func (proj *Project) Warnings() []string {
...
+			return true, nil
+		}
+
+		if strings.HasPrefix(trimmed, "n") {
+			// User wants to cancel.
+			return false, nil
 		}
+
+		// Invalid response.
+		fmt.Printf("Invalid response.  Please enter 'y' or 'n'.\n")
```
utzig: A small nit! `Please enter 'y' or 'n'` since you already have a prompt with `[Y/n]`, and both `<ENTER>`, `YES`, `Yes` would work this extra message feels unnecessary (and not precise).

newt/downloader/downloader.go
```diff
@@ -667,8 +667,8 @@ func (ld *LocalDownloader) CurrentBranch(path string) (string, error) {
 }
 
 func (ld *LocalDownloader) UpdateRepo(path string, branchName string) error {
-	// Nothing to update in a local repo.
-	return nil
+	os.RemoveAll(path)
```
utzig: Hmmm, if I understand this correctly, it would be completely removing the local path including any changes that the user might have made, is that ok?

newt/repo/repo.go
```diff
@@ -596,18 +499,18 @@ func (r *Repo) Sync(vers *Version, force bool) (bool, error) {
 		msg := "Unexpected local branch for %s: \"%s\" != \"%s\""
 		if force {
 			util.StatusMessage(util.VERBOSITY_DEFAULT,
-				msg+"\n", r.rdesc.name, currBranch, branchName)
+				msg+"\n", r.Name(), currBranch, branchName)
 		} else {
-			err = util.FmtNewtError(
-				msg, r.rdesc.name, currBranch, branchName)
+			err := util.FmtNewtError(
+				msg, r.Name(), currBranch, branchName)
```
utzig: hmmmm, I thought the rule for short variable declaration was that at least one variable must be newly declared... but `err` is already declared above...

utzig:
Could you break this into multiple PRs that build feature by feature? Just kidding, the code looks really good and the comments are a very nice addition. I will give it some testing locally but I would not oppose merging this now...

utzig:
Well, a `rm -rf bin && for i in targets/*; do newt build $(echo $i | sed 's#targets/##g'); done` works here so no regressions on my tens of targets :P

newt/repo/repo.go
ccollins476ad: Since this `err` is declared in a child scope, it shadows the earlier declaration.  This is analogous to the C code:

```
int err = 10;
while (keep_going()) {
    int err = 20;
    /* err == 20 */
}
/* err == 10 */
```

I think shadowing a variable declaration can be confusing, and is probably best avoided.  In the case of an `err` variable, I don't think there is that much room for confusion, however.  Though, in this case, I don't think the second `err` variable should be declared at all, so I am going to change it to:

```
            return false, util.FmtNewtError(
                msg, r.Name(), currBranch, branchName)
```

newt/downloader/downloader.go
ccollins476ad: It isn't ideal.  Ideally, newt would detect changes by comparing all files against the local repo's path.

The `local` repo type is just a debug feature, so I was a bit sloppy here.  I think a fix for this might be good as a separate PR.  What do you think?

newt/project/project.go
ccollins476ad: I agree - the specifics in the message are unnecessary.  I will change it to:
```
Invalid response.
```

newt/repo/repo.go
ccollins476ad: That's pretty funny :).  Thanks, I have fixed it.

newt/downloader/downloader.go
utzig: Yes, that would be ok... I just pointed out because I was not sure, thanks!

PR #148, opened 2018-03-14
Forced reinstall (`install -f`) was always failing

ccollins476ad:
Forced reinstalls always failed with an error message similar to the following:
```
Error: Error updating "blehostd": chdir /Users/ccollins/tmp/myproj2/repos/blehostd: no such file or directory
```

The problem occurred when newt attempted to fetch the repo after deleting the repo's directory.

The fix is to re-clone the repo rather than fetch it if the directory is missing.

### DIFF
--- a/newt/repo/repo.go
+++ b/newt/repo/repo.go
@@ -219,12 +219,19 @@ func (r *Repo) checkExists() bool {
 }
 
 func (r *Repo) updateRepo(branchName string) error {
-	dl := r.downloader
-	err := dl.UpdateRepo(r.Path(), branchName)
+	err := r.downloader.UpdateRepo(r.Path(), branchName)
 	if err != nil {
-		return util.FmtNewtError("Error updating \"%s\": %s",
-			r.Name(), err.Error())
+		// If the update failed because the repo directory has been deleted,
+		// clone the repo again.
+		if util.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


newt/repo/repo.go
```diff
@@ -219,12 +219,19 @@ func (r *Repo) checkExists() bool {
...
 	if err != nil {
-		return util.FmtNewtError("Error updating \"%s\": %s",
-			r.Name(), err.Error())
+		// If the update failed because the repo directory has been deleted,
+		// clone the repo again.
+		if util.IsNotExist(err) {
+			err = r.downloadRepo(branchName)
+		}
 	}
+	if err != nil {
```
mkiiskila: Lines 229-230 could be removed.

utzig: Just removing those lines slightly changes the behaviour... following the path of `err != nil` in line 223 might run `downloadRepo` and return `err == nil` in which case it would end up running `util.FmtNewError`...

mkiiskila: True. My mistake in comment :) I meant line 229 only; no need to check err == nil twice.

sterlinghughes:
can we merge this?

ccollins476ad:
Thanks @mkiiskila and @utzig.  I have removed the redundant error check.

PR #152, opened 2018-04-01
builder/cmake: Native linker fixes

ynezz:
While exploring `newt`, I've hit some issues with `cmake` builder. This PR fixed the issues for me.


sterlinghughes:
@michal-narajowski can you look at this in context of your DNM PR for cmake?  Let's see if we can merge one or both of these PRs soonish so that cmake remains working.

PR #161, opened 2018-04-18
cmake: Use paths relative to project root

michal-narajowski:
Absolute paths shouldnt be necessary, CMake takes care of
paths. Some users had issues with using absolute paths.
This makes the CMakeLists.txt file smaller.

### DIFF
--- a/newt/builder/cmake.go
+++ b/newt/builder/cmake.go
@@ -30,6 +30,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"mynewt.apache.org/newt/newt/interfaces"
 	"mynewt.apache.org/newt/newt/project"
 	"mynewt.apache.org/newt/newt/target"
 	"mynewt.apache.org/newt/newt/toolchain"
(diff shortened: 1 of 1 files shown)


mwilliford:
+1 works like a champ!

newt/builder/cmake.go
```diff
@@ -46,6 +47,19 @@ func EscapeName(name string) string {
...
 
+func trimProjectPath(path string) string {
+	proj := interfaces.GetProject()
+	path = strings.TrimPrefix(path, proj.Path()+"/")
+	return path
+}
+
+func trimProjectPathSlice(elements []string) []string {
+	for e := range elements {
+		elements[e] = trimProjectPath(elements[e])
```
ccollins476ad: I would be a bit careful with this.  In Go, slices are passed by reference, so this line is actually modifying the slice that the caller passed in.  It might be better to do one of the following:
1. Return a new slice containing the trimmed elements, or
2. Change the function type such that it doesn't return anything.

From the function type, a user might assume the function doesn't modify his original element slice.

michal-narajowski:
Thanks, @ccollins476ad, good catch. Fixed!

PR #163, opened 2018-04-20
Fix conditional dependency lists

ccollins476ad:
This fixes https://github.com/apache/mynewt-newt/issues/162.  Please see the issue for more information.

This is a rather large PR, but it fixes the issue, and I believe it makes the newt code clearer.  The fix involves these two changes:

1. Delay API requirement resolution until after dependency resolution and syscfg calculation.  This is done to prevent an incorrect package from being selected as an API-supplier before dependency resolution has completed.
2. Allow a package to be deleted from the syscfg state.  When a package is invalidated by a conditional dependency list, all setting definitions and overrides belonging to the package need to be removed from the system.

### DIFF
--- a/newt/resolve/resolve.go
+++ b/newt/resolve/resolve.go
@@ -24,8 +24,6 @@ import (
 	"sort"
 	"strings"
 
-	log "github.com/Sirupsen/logrus"
-
 	"mynewt.apache.org/newt/newt/flash"
 	"mynewt.apache.org/newt/newt/pkg"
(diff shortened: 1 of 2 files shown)


ccollins476ad:
FYI- I introduced a bug, so I combined the fix with an existing commit and force pushed.  The bug was: newt didn't complain when an API requirement wasn't satisfied.  Instead, the build would proceed until a compile or link error was encountered.

PR #164, opened 2018-04-24
Add support for declaring a key file under target

utzig:
This allows the user to input a key file using `target.public_key`. The file will be parsed and an output .c file with the bootloader required public key will be written.


utzig:
This implements the suggestion of this comment: https://github.com/runtimeco/mcuboot/issues/203#issuecomment-378239740

I did basic testing only, needs better error checking, need to check it did not break image signing.

utzig:
Also, `public_key` sounds like a very bad name for something that can be a PEM private key as well...

mkiiskila:
More natural place for the generated file would be in the target build directory, the same spot where sysinit generated c-file is placed. Aka bin/targets/<tgt_name>/generated/src
Did you consider this idea already?

utzig:
@mkiiskila I have not, but that seems like the best approach.

utzig:
Considering that one would usually avoid sharing a private key (unless in development), I updated this so that public keys in PEM format are also accepted. Instructions to generate were added as comments and `signed_images.md` can be updated after this is merged. The keys can currently be specified with an absolute path, otherwise it is relative to the root of `mynewt-core` tree. Not sure this would be the preferred behaviour.

Btw, tested this with all combinations of RSA and EC256 keys. Still need to test EC224 (if that is still supported at all).

Also some `TODO` / `FIXME` need checking or removal.

PR #169, opened 2018-05-23
Improve size command  output

sjanc:
This allows to handle multiple repos as well as application code
correctly (by adding required nesting).


newt/builder/size.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	"mynewt.apache.org/newt/newt/image"
 	"mynewt.apache.org/newt/util"
+        "mynewt.apache.org/newt/newt/interfaces"
```
utzig: change spaces -> tab

sjanc: fixed

PR #172, opened 2018-05-31
Always refer to packages by their full name

### LINKED ISSUE #171: Unable to have same named packages in two separate repos
newt build is not taking repository when checking for package name

eg we have two samples
apache-mynewt-nimble/apps/btshell
apache-mynewt-core/apps/btshell

and then target
```
[janc@ix mynewt]$ newt target show btshell
targets/btshell
    app=@apache-mynewt-core/apps/btshell
    bsp=@apache-mynewt-core/hw/bsp/nrf52840pdk
```

It is not possible to build this:
[janc@ix mynewt]$ newt build btshell
Building target targets/btshell
Error: Two app packages in build: apps/btshell, apps/btshell

I think newt should be able to distinguish packages with same name but from different repos. Especially now that we have nimble in separate repo (and more to come).

ccollins476ad:
This fixes #171.

Packages have two forms of "name":
    * Name()
    * FullName()

`Name()` is everything except the repo prefix ("@<repo-name>").  Full name includes the repo prefix.  The normal `Name` was used in several places, causing ambiguity when two repos contained packages with the same name.

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -92,11 +92,11 @@ func NewBuilder(
 		bpkg := b.PkgMap[rpkg]
 		if bpkg == nil {
 			for _, rpkg := range b.sortedRpkgs() {
-				log.Debugf("    * %s", rpkg.Lpkg.Name())
+				log.Debugf("    * %s", rpkg.Lpkg.FullName())
 			}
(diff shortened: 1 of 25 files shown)


vrahane:
Looks good. Tested it out.

PR #177, opened 2018-06-14
 Add support for additional comparison operators 

ccollins476ad:
This commit adds support for four additional comparison operators in newt expressions: `<`, `<=`, `>`, `>=`.

Both operands must be numbers.  If one or more operands is not a number, newt raises an error.

### DIFF
--- a/newt/parse/lex.go
+++ b/newt/parse/lex.go
@@ -32,6 +32,10 @@ const (
 	TOKEN_NOT_EQUALS TokenCode = iota
 	TOKEN_NOT
 	TOKEN_EQUALS
+	TOKEN_LT
+	TOKEN_LTE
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
Could you link to an example of usage of of these in the newt code base, not related to this PR, just in general.

ccollins476ad:
@jacobrosenthal, there is no example usage in the newt code base because the feature hasn't been merged yet :).  If you just want to see an example: say you implement a feature which requires a lot of mbuf space:
```
syscfg.defs:
    MY_FEATURE:
        value: 1
        restrictions:
            - 'MSYS_1_BLOCK_COUNT >= 16 && MSYS_1_BLOCK_SIZE >= 400'
```

PR #183, opened 2018-06-21
141

sjanc:


utzig:
After seeing this it makes more sense to me calling branches just `1_3_dev`, `1_4_dev` and so on...

PR #185, opened 2018-06-23
Sync: Pull from current upstream, not origin

ccollins476ad:
Prior to this commit, the `newt sync` command would assume the current branch's upstream is origin.  This caused a number of issues:

1. If branch has an upstream other than origin, but there exists an identically named branch in origin, then `origin/<branch>` is pulled into the current checkout.  So, updates to one branch would get merged into another, i.e., totally wrong.

2. If origin does not have an identically named branch, then the sync operation would have no effect.

This commit changes newt to just pull from the currently tracked upstream.

### DIFF
--- a/newt/cli/project_cmds.go
+++ b/newt/cli/project_cmds.go
@@ -58,7 +58,7 @@ func newRunCmd(cmd *cobra.Command, args []string) {
 	}
 	defer os.RemoveAll(tmpdir)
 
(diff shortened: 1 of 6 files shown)


ccollins476ad:
Note: this PR looks enormous due to the "revendor" commit.  The revendor commit can probably be ignored during review.

mkiiskila:
Could this be rebased?

ccollins476ad:
Thanks for the heads up.  I have fixed the conflicts.

PR #190, opened 2018-07-07
Convert from godep to dep for vendoring

ccollins476ad:
This PR switches the newt repo over to the official `dep` vendor tool.  This repo was previously using the old `godep` tool.

Aside from being "official" and better maintained, `dep` has a few advantages over `godep`:
* The `vendor` directory is a top-level directory, not a subdirectory of `newt/newt`.  This means that `newt`, `util`, `viper`, and all other Go projects in this repo can share the same vendor directory.  It also means that each Go project doesn't need to store a second copy of its sibling projects in its `vendor` directory.

* Bug fix: OS-specific vendored packages are allowed.  With Godep, these caused `go get` to abort with an error when run on a different OS.


utzig:
LGTM

PR #191, opened 2018-07-10
Cmake fixes

michal-narajowski:
- cmake: Escape quotes in linker flags

- cmake: Fix undeterministic build error by sorting includes
The order of include directories was non-deterministic and caused issues with
including wrong files which resulted in 'undefined reference' errors.

### DIFF
--- a/newt/builder/cmake.go
+++ b/newt/builder/cmake.go
@@ -229,7 +229,7 @@ func (b *Builder) CMakeTargetWrite(w io.Writer, targetCompiler *toolchain.Compil
 		elfOutputDir,
 		elfOutputDir,
 		elfOutputDir,
-		strings.Join(lFlags, " "))
+		strings.Replace(strings.Join(lFlags, " "), "\"", "\\\\\\\"", -1))
 
 	fmt.Fprintln(w)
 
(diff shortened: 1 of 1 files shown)


newt/builder/cmake.go
```diff
@@ -263,7 +244,7 @@ func CmakeCompilerInfoWrite(w io.Writer, archiveFile string, bpkg *BuildPackage,
 	includes = append(includes, c.GetLocalCompilerInfo().Includes...)
 	includes = append(includes, otherIncludes...)
 
-	includes = removeDuplicates(includes)
+	includes = util.SortFields(includes...)
```
utzig: Is it not a problem to leave duplicates?

michal-narajowski: This function also removes duplicates. It is used by newt when compiling, so it should be ok.

utzig: Ah right, `SortFields` calls `UniqueStrings` but that is not documented behavior!

michal-narajowski:
I added a comment that explains what sortFields() does and added a commit that sorts and removes duplicate compilation flags to account for possible issues with ordering in this aspect.

PR #195, opened 2018-07-17
Allow build_profile to be overridden per package

ccollins476ad:
The build profile can be overridden for a particular pacakge in two
ways:

* (1) Specify a "pkg.build_profile" value in the package's `pkg.yml` file.  E.g.,
```
pkg.build_profile: debug
```

* (2) Specify the pacakge in the target's "target.package_profiles" map.  E.g.,
```
target.package_profiles:
    'apps/blinky': debug
    '@apache-mynewt-core/sys/log/full': debug
```

(1) takes priority over (2).

In addition, it is now possible to override the optimization setting ("-O<x>") in a package's `pkg.cflags`.  Prior to this change, the "-O" setting specified by the compiler / build profile would always override what other packages specify.

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -245,10 +245,38 @@ func collectCompileEntriesDir(srcDir string, c *toolchain.Compiler,
 	return entries, nil
 }
 
+// Determines which build profile to use when building the specified package.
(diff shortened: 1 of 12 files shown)


ccollins476ad:
Thanks, @utzig, those are all good comments.  I have made the recommended changes.

PR #197, opened 2018-07-25
syscfg: error detection after API resolution

ccollins476ad:
Prior to this PR, newt detected errors in the syscfg before API requirements were satisfied. Consequently, error detection was happening before the full dependency graph had been generated, resulting in an incomplete set of syscfg settings.  Parsing of restriction strings might fail in this case, as unrecognized setting names are parsed as string literals rather than expanded to the value that the setting would ultimately have.

This PR delays error detection until after API resolution.

### DIFF
--- a/newt/syscfg/syscfg.go
+++ b/newt/syscfg/syscfg.go
@@ -1009,6 +1009,15 @@ func (cfg *Cfg) detectAmbiguities() {
 	}
 }
 
+// Detects and records errors in the build's syscfg.  This should only be
+// called after APIs are resolved to avoid false positives.
+func (cfg *Cfg) DetectErrors(flashMap flash.FlashMap) {
+	cfg.detectAmbiguities()
+	cfg.detectViolations()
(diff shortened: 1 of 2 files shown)


newt/syscfg/syscfg.go
```diff
@@ -1009,6 +1009,15 @@ func (cfg *Cfg) detectAmbiguities() {
 	}
 }
 
+// Detects and records errors in the build's syscfg.  This should only be
+// called after APIs are resolved to avoid false positives.
+func (cfg *Cfg) DetectErrors(flashMap flash.FlashMap) {
```
utzig: Wouldn't it be better to receive `flashMap` as pointer?

ccollins476ad: Yeah, that's a good question.  I never really know whether to pass by value or pointer in Go.  These days I tend to pass by value unless the function is intended to modify the data since there is no `const`.

PR #201, opened 2018-08-21
Allow setting value to reference other setting

andrzej-kaczmarek:
Currently the only way to have one setting copy value from other setting is to do as follows:
```
syscfg.vals:
    FOO: 'MYNEWT_VAL_BAR'
```
The problem with the above is that `newt` handles value of `FOO` as a string which is then resolved by preprocessor during build so this won't evaluate properly if used inside `syscfg.restrictions`

This PR allows one setting value to explicitly reference other setting and have other value copied by `newt`. The syntax is the same as used in C code:
```
syscfg.vals:
    FOO: 'MYNEWT_VAL(BAR)'
````

For now it does not support recursive references but I think we can live with it until I can figure out how to do this in Go ;-)

### DIFF
--- a/newt/cli/target_cmds.go
+++ b/newt/cli/target_cmds.go
@@ -567,17 +567,34 @@ func printSetting(entry syscfg.CfgEntry) {
 		util.StatusMessage(util.VERBOSITY_DEFAULT,
 			"default=%s\n", entry.History[0].Value)
 	}
+	if len(entry.ValueRefName) > 0 {
+		util.StatusMessage(util.VERBOSITY_DEFAULT,
+			"    * Copied from: %s\n",
+			entry.ValueRefName)
+	}
 }
 
 func printBriefSetting(entry syscfg.CfgEntry) {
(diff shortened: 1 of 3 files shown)


ccollins476ad:
I like the feature, but I have two comments:

1. I am not sure how I feel about `@` as a prefix.  Since this is already used to designate a repo name, I worry this could be confusing for users.  I wonder if it would be clearer if the notation were: `MYNEWT_VAL(setting_name)`.  I am really not sure, though... if you think `@` is fine then it is probably OK.

2. Would you mind renaming `ResolveValues()` to something less generic?  Maybe `ResolveReferences()`?

andrzej-kaczmarek:
Thanks, Chris!
Tbh `@` was just first prefix that came to my mind and I did not think it through. I agree it may be confusing since it's already use for other purpose so I changed syntax to `MYNEWT_VAL(xxx)` which feels right as it follows C code. Also method and variable names were adjusted in new version.

PR #202, opened 2018-08-24
Add new package type -> transient

andrzej-kaczmarek:
**The problem:** we want to move/rename some packages in a backwards-compatible manner.
For `lib` packages we can just move everything to new package and make simple dependency using `pkg.deps` from old (empty) package to new one. This works fine and was used for NimBLE migrations. The only drawback is that `newt` does not warn about the rename.
However, the same approach cannot be used for `bsp` or `app` (and few other types) packages since they are read directly as specified in target settings so cannot be just dummy packages.

**The solution:** use transient packages
Transient package is dummy (empty) package which only link other package. Unlike dependencies it can link only single package. They can be specified in target settings since target package will be used automatically. Other than this, link is resolved just like regular dependency.

To define transient package it's enough to create simple `pkg.yml` file as follows:
```
pkg.name: old-package
pkg.type: transient
pkg.link: @repo/target-package
```

Including transient package in a project will always emit a warning:
```
andk@t480s:~/devel/mynewt$ newt build btshell 
Building target targets/btshell
2018/08/24 21:36:46.229 [WARNING] Transient package @apache-mynewt-core/hw/bsp/nrf52840pdk used, update configuration to use target package instead (@apache-mynewt-core/hw/bsp/nordic_pca10056)
Target successfully built: targets/btshell
```

Some existing PRs using transient packages to test with:
https://github.com/apache/mynewt-core/pull/1346
https://github.com/apache/mynewt-core/pull/1348

### DIFF
--- a/newt/resolve/resolve.go
+++ b/newt/resolve/resolve.go
@@ -661,7 +661,7 @@ func ResolveFull(
 		}
 
 		log.Warnf("Transient package %s used, update configuration "+
-			"to use target package instead (%s)",
+			"to use linked package instead (%s)",
 			rpkg.Lpkg.FullName(), rpkg.Lpkg.LinkedName())
 	}
 
--- a/newt/resolve/resolve.go
+++ b/newt/resolve/resolve.go
@@ -24,11 +24,14 @@ import (
 	"sort"
 	"strings"
 
+	log "github.com/Sirupsen/logrus"
+
 	"mynewt.apache.org/newt/newt/flash"
 	"mynewt.apache.org/newt/newt/pkg"
 	"mynewt.apache.org/newt/newt/project"
 	"mynewt.apache.org/newt/newt/syscfg"
 	"mynewt.apache.org/newt/util"
+	"mynewt.apache.org/newt/newt/ycfg"
 )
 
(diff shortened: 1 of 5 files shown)


newt/resolve/resolve.go
```diff
@@ -644,6 +653,18 @@ func ResolveFull(
...
+	// We have now resolved all packages so go through them and emit warning
+	// when using link packages
+	for _, rpkg := range res.MasterSet.Rpkgs {
+		if rpkg.Lpkg.Type() != pkg.PACKAGE_TYPE_TRANSIENT {
+			continue
+		}
+
+		log.Warnf("Transient package %s used, update configuration "+
+			"to use target package instead (%s)",
+			rpkg.Lpkg.FullName(), rpkg.Lpkg.LinkedName())
```
ccollins476ad: This warning might be confusing since "target" already means something else.  Maybe call it the "linked package" instead of "target package"?

andrzej-kaczmarek:
Thanks Chris!
Agree about warning message, I pushed separate commit to fix it and will merge this PR now since it's approved already.

PR #211, opened 2018-09-05
[WIP] Follow Newt compatibility list when doing newt install and upgrade

michal-narajowski:


michal-narajowski:
This was surpassed by https://github.com/apache/mynewt-newt/pull/219. Closing this one.

PR #212, opened 2018-09-06
Add support for encrypted images

utzig:
This adds a new flag to `create-image` command: `-e <rsa-pub-key.pem>` to allow generation of encrypted images. This adds a layer of secrecy while transporting images to the device and for storing at an external flash.

To encrypt, RSA-OAEP was used, mainly because it is well supported by libraries. This requires the device to already have the private key stored internally. The encryption process method follows these steps:

1) Generate a random AES-128 key
2) Use this key with AES-128-CTR to encrypt the image
3) Use RSA-OAEP to encrypt the key
4) Store this encrypted key in a new TLV (0x30)
5) A new header flag signals the image to be encrypted.

When doing a swap, the image is decrypted from slot1 -> slot0, and re-encrypted when coming back from slot0 -> slot1 based on the header flag.

The TLV adds about 256 bytes to the image, and RSA-OAEP in the firmware increases some 12KB of flash required, so better key encryption methods might be added later, like EICES (using eliptic curves) and AES-KEK using a secret key to encrypt the encryption key itself.

This is part of https://github.com/runtimeco/mcuboot/pull/330


utzig:
I think this is not going to change, but `mcuboot` is still pending a few features.

utzig:
With latest changes now `-e` also accept an AES-128 symmetric key generated with:

`$ dd if=/dev/urandom of=/dev/stdout bs=1 count=16 | base64 > aes-128-kek.b64`

PR #215, opened 2018-09-24
Fix usage of -f on target delete/init commands

utzig:


utzig:
using `delete -f` is not functional and blocking the `blinky` build tests on travis...

PR #221, opened 2018-10-16
logcfg: Allow logs to be defined in `syscfg.yml`

ccollins476ad:
This PR adds a new feature to newt: A package can define logs in its `syscfg.yml` file under the heading of `syscfg.logs`.  During the build process, newt generates a C header file called `logcfg/logcfg.h` containing macros for using each generated log.  This feature is described in more detail in the commit comments.

Please feel free to critique the feature itself in addition to its implementation.

This feature serves two purposes:
1. Allow finer-grained control of which log messages to include in the build.
2. Provide a means of visualizing the various log modules in the system.

#### Allow finer-grained control of which log messages to include in the build
Currently, the user can control which log messages get included in the build by setting the `LOG_LEVEL` syscfg setting.  However, this setting applies to every log in the system, and may be too coarse to be useful.  This PR allows the log level to be set per module rather than system-wide.

#### Provide a means of visualizing the various log modules in the system
Module IDs throughout the system must be unique.  Since third-party packages may define their own log modules, it is currently quite difficult to enforce this constraint.  This PR introduces a few newt commands (`target log show` and `target log brief`) that display tables describing the log modules in use by the system.  Moreover, newt now aborts the build if it detects a log module conflict.

NOTE: There is a small sibling PR for mynewt-core: https://github.com/apache/mynewt-core/pull/1461

### DIFF
--- a/newt/builder/targetbuild.go
+++ b/newt/builder/targetbuild.go
@@ -101,6 +101,9 @@ func NewTargetTester(target *target.Target,
 		injectedSettings: map[string]string{},
 	}
 
+	// Indicate that this version of newt supports the generated logcfg header.
+	t.InjectSetting("NEWT_FEATURE_LOGCFG", "1")
+
 	return t, nil
 }
 
@@ -214,6 +217,12 @@ func (t *TargetBuilder) validateAndWriteCfg() error {
 	}
 
 	if err := syscfg.EnsureWritten(t.res.Cfg,
+		GeneratedIncludeDir(t.target.Name())); err != nil {
+
+		return err
+	}
+
+	if err := t.res.LCfg.EnsureWritten(
 		GeneratedIncludeDir(t.target.Name())); err != nil {
 
 		return err
--- a/newt/cli/target_cmds.go
+++ b/newt/cli/target_cmds.go
@@ -31,6 +31,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"mynewt.apache.org/newt/newt/builder"
+	"mynewt.apache.org/newt/newt/logcfg"
 	"mynewt.apache.org/newt/newt/newtutil"
 	"mynewt.apache.org/newt/newt/pkg"
 	"mynewt.apache.org/newt/newt/resolve"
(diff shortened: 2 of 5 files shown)


andrzej-kaczmarek:
This is a really nice idea, hopefully someone can review Go code and we can merge this soon :-)

Just few things to consider:
- what about some "don't care" (`-1`?) module id which means "first available value" (from some predefined range)?
- this could also generate file with `id` -> `name` mappings so these can be returned by e.g. `newtmgr` and perhaps also used for some pretty-printing later
- I'm not sure about the `guest` field. perhaps if we want to reuse id of some other module we could have some other field to reference this by name? my concern is that if we change module id of "host" module then it needs to be updated by all "guests"... if we reference by name, this is not a problem.

and I guess eventually we'll be able to remove hardcoded values from `log_common.h` and replace with this new feature?

ccollins476ad:
> Just few things to consider:
> 
> * what about some "don't care" (`-1`?) module id which means "first available value" (from some predefined range)?

Good idea.

> * this could also generate file with `id` -> `name` mappings so these can be returned by e.g. `newtmgr` and perhaps also used for some pretty-printing later

I think this is good also.  I was actually thinking something similar after I submitted this PR.  This would remove the need to manually register module names in the application.

> * I'm not sure about the `guest` field. 

I am not sure about `guest` either.  I don't remember what I was thinking exactly.  I think there may be the desire for two different packages to log to the same module, but both packages need their own macros so that they can be disabled independently at build time.  This may not be a real case after all.  I am fine with removing `guest` to simplify the feature.  We can re-add it later if necessary.

> perhaps if we want to reuse id of some other module we could have some other field to reference this by name? my concern is that if we change module id of "host" module then it needs to be updated by all "guests"... if we reference by name, this is not a problem.

We can already reference a module by name if there is a dedicated syscfg setting (e.g., `PKG1_LOG_MODULE`).  If PKG2 wants to use PKG1's log module, it can just specify `MYNEWT_VAL(PKG1_LOG_MODULE)` as its module ID.  This requires a bit of extra work; each package needs to define syscfg settings for its module IDs, but I think this is desirable, as that lets the target override all module IDs in the system.

> and I guess eventually we'll be able to remove hardcoded values from `log_common.h` and replace with this new feature?

Yes, I think that will be a good thing.  We can also remove the distinction between "system modules" and "user modules" that exists today (system module IDs are in the range 0-63).

ccollins476ad:
Closing; this was superseded by #230.

PR #237, opened 2018-11-08
Merge conflicting dependency conditionals

ccollins476ad:
**NOTE:** This PR includes #235.  That should be merged before this one.

This PR addresses the following example:
```
    pkg.deps.FOO:
        - my_pkg
    pkg.deps.BAR:
        - my_pkg
```

This package has two conflicting conditional dependencies.

Before this PR, newt would only process one of the dependencies, and it did so in an inconsistent order.  So, the build would end up with one or the other of the dependencies, but never both.

Now, newt merges such conflicting conditional dependencies into a single dependency.  The resulting dependency is conditional on the union of all the sub expreessions.  Newt would (internally) convert the above example to:                  
```
    pkg.deps.'(FOO) || (BAR)'
        - my_pkg
```

Alternatively, the user could perform this transformation manually, but that could get messy with complicated `pkg.yml` files.

### DIFF
--- a/newt/target/target.go
+++ b/newt/target/target.go
@@ -20,10 +20,10 @@
 package target
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	"mynewt.apache.org/newt/newt/newtutil"
 	"mynewt.apache.org/newt/newt/pkg"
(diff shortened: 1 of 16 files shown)


newt/target/target.go
```diff
@@ -56,9 +56,13 @@ type Target struct {
 
 func NewTarget(basePkg *pkg.LocalPackage) *Target {
 	target := &Target{
-		TargetY: ycfg.YCfg{},
+		basePkg: basePkg,
 	}
-	target.Init(basePkg)
+
+	if basePkg.SyscfgY.Tree() == nil {
+		panic("ASD")
```
mkiiskila: While "ASD" seems great, maybe something more descriptive? ;)

ccollins476ad: I really don't see what the problem is.  JK- I don't know how I missed that, thanks!

PR #239, opened 2018-11-14
docs; give examples about how to use extrajtagcmd to debug multiple b…

mkiiskila:
…oards simultaneously.

### DIFF
--- a/docs/newt_ops.rst
+++ b/docs/newt_ops.rst
@@ -67,3 +67,13 @@ through the ``-h`` or ``--help`` options.
           -v, --verbose           Enable verbose output when executing commands
 
         Use "newt target [command] --help" for more information about a command.
+
+
+It is possible to use *newt* to debug multiple boards simultaneously. To do that, you'll need to select TCP port gdb uses to talk with JTAG adapter software, and also pick the JTAG adapter to use. Selected TCP ports should be unique, and available for use. These can be passed within parameter ``-extrajtagcmd`` wnen invoking ``newt load``, ``newt debug`` or ``newt run``.
+
+Here are few examples:
+
+.. code-block:: console
+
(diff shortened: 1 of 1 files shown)


docs/newt_ops.rst
```diff
@@ -67,3 +67,13 @@ through the ``-h`` or ``--help`` options.
...
+
+
+It is possible to use *newt* to debug multiple boards simultaneously. To do that, you'll need to select TCP port gdb uses to talk with JTAG adapter software, and also pick the JTAG adapter to use. Selected TCP ports should be unique, and available for use. These can be passed within parameter ``-extrajtagcmd`` wnen invoking ``newt load``, ``newt debug`` or ``newt run``.
+
+Here are few examples:
+
+.. code-block:: console
+
+        newt debug blinky_nrf52 --extrajtagcmd="-port 3338 -select usb=682223238"
+	newt run slinky_zero 4.0 --extrajtagcmd="cmsis_dap_serial 00000000AZE000001422 -port 3400"
```
saulpw: Note that `<port>` is the gdb port, but `<port>+1` is also used as the telnet port, making it unavailable.  This bit me when I chose consecutive ports for the two sessions.

I also had to add ` ; tcl_port <n>` and specify a third port, which is not auto-specified by newt like the telnet port is, and will collide if not specified (defaults to 5555).  Note also the spaces around the `;`, which seem to be necessary for openocd.

Finally, maybe a mention of how to get the serial of the device?  On Linux, I used `lsusb -v` and found the `iSerial` property on the adapter.

aditihilbert: @mkiiskila Do you want to incorporate the suggestions before we merge? Thanks!

ccollins476ad:
These changes are outdated I am closing this.

PR #247, opened 2018-11-26
Include git hash and build date in version output

ccollins476ad:
The hash and build date need to be specified at build time.  This has been added to the `build.sh` script.


ccollins476ad:
Thanks, @mkiiskila.  That is a good idea.

PR #253, opened 2018-12-12
Fix table syntax

utzig:


utzig:
This just fixes syntax of a table in the RestructuredText which is currently broken and not visible by the users. I can updated whatever docs are need in another PR. Also brew install actually calls brew update so there's no reason to call manually but the results should be the same anyway.

PR #254, opened 2018-12-28
2.0 mfgimage format

ccollins476ad:
This PR implements the "new" manufacturing image format as discussed here: https://lists.apache.org/thread.html/4185891ef587a778722a39fe5a1ae78bfd0dc7ab7e9348a7ea147df7@%3Cdev.mynewt.apache.org%3E

This PR also includes two other PRs:
* (artifact library) https://github.com/apache/mynewt-newt/pull/243
* (larva tool) https://github.com/apache/mynewt-newt/pull/244


artifact/image/key.go
```diff
@@ -265,35 +294,48 @@ func parseEncKeyBase64(keyBytes []byte, plainSecret []byte) ([]byte, error) {
...
+	if _, err := rand.Read(plainSecret); err != nil {
 		return nil, util.FmtNewtError(
-			"Error reading pubkey file: %s", err.Error())
+			"Random generation error: %s\n", err)
 	}
 
+	return plainSecret, nil
+}
+
+func GenerateCipherSecret(pubKeBytes []byte,
```
utzig: Any reason for abbreviating `Key` to `Ke` in this function? Using just `PubK...` would be better IMO.

ccollins476ad: Thanks, @utzig!  `Ke` is short for key-encrypting.  This is the key-encrypting key; it is a key that encrypts the actual secret that encrypts the image.

utzig: It should be called `Kek` than! ;-)

PR #258, opened 2019-01-10
larva: Replace `swapkey` with `swapisk`, `swapkek`

ccollins476ad:
The two commands are necessary because the KEK length field needs to be updated as well.  The ISK (image signing key) is a public key, so its length is constant.

### DIFF
--- a/artifact/image/keys_test.go
+++ b/artifact/image/keys_test.go
@@ -22,8 +22,10 @@ package image_test
 import (
 	"io/ioutil"
 	"os"
-	"path"
 	"testing"
+
+	"mynewt.apache.org/newt/artifact/image"
+	"mynewt.apache.org/newt/artifact/sec"
 )
 
 func TestRSA(t *testing.T) {
@@ -43,15 +45,15 @@ func TestPlainEcdsaPkcs8(t *testing.T) {
 }
 
 func TestEncryptedRSA(t *testing.T) {
-	image.KeyPassword = []byte("sample")
+	sec.KeyPassword = []byte("sample")
 	signatureTest(t, rsaEncryptedPrivate)
-	image.KeyPassword = []byte{}
+	sec.KeyPassword = []byte{}
 }
 
 func TestEncryptedEcdsa(t *testing.T) {
-	image.KeyPassword = []byte("sample")
+	sec.KeyPassword = []byte("sample")
 	signatureTest(t, ecdsaEncryptedPrivate)
-	image.KeyPassword = []byte{}
+	sec.KeyPassword = []byte{}
 }
 
 func signatureTest(t *testing.T, privateKey []byte) {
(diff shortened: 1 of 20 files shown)


artifact/image/keys_test.go
```diff
@@ -65,58 +67,29 @@ func signatureTest(t *testing.T, privateKey []byte) {
...
-	tmp := make([]byte, 256)
-	for i := 0; i < len(tmp); i++ {
-		tmp[i] = byte(i & 0xFF)
-	}
-	err = ioutil.WriteFile(simpleName, tmp, 0644)
-	if err != nil {
-		t.Fatal(err)
+	body := make([]byte, 256)
+	for i := 0; i < len(body); i++ {
+		body[i] = byte(i & 0xFF)
```
utzig: Why is doing an `&` here different than just doing `= byte(i)`?

larva/lvmfg/lvmfg.go
```diff
@@ -160,32 +161,77 @@ func Join(mm NameBlobMap, eraseVal byte,
...
+	if _, err := replaceKey(mfgBin, okey, nkey); err != nil {
+		return err
 	}
 
-	if len(okey) > len(sec0) {
+	return nil
+}
+
+func ReplaceKek(mfgBin []byte, okey []byte, nkey []byte) error {
+	if len(nkey) > len(okey) {
```
utzig: So if `nkey` cannot be larger than `okey` what is the purpose of having the size encoded before the key? Suppose `nkey` is larger than `okey`, then I pass `nkey` first, `okey` next (it's smaller so this is OK), and then pass `nkey` again, same as original key but this time it's larger than the one currently there; isn't this supposed to work? (I might be missing something most probably!)

artifact/image/keys_test.go
ccollins476ad: Good spot.  Casting to `byte` will perform the desired truncation automatically.

larva/lvmfg/lvmfg.go
ccollins476ad: This is indeed confusing.  I am sure this could use some explanatory comments or documentation.

First, recall what this command does - it replaces the KEK in a boot loader binary with a new one.  The larva tool has no control over how the KEK is encoded in the binary.  This tool must work with the existing format.

In the boot loader, the KEK is stored as a pair of values:
* ASN.1 encoded ("DER") private key
* 32-bit length of the key, in bytes

For larva to replace the existing key, it must update both of the above values.  This is the equivalent of brain surgery, so it is necessary to ensure nothing goes wrong during the operation.  To this end, larva imposes the following restrictions:

1. The new key must not be longer than the old key (if it were, the new key would extend beyond the key buffer and into other code).
2. The key length variable must be placed immediately prior to the key itself.

(The second restriction is easy to arrange: place both items in a struct with the length as the first member.)

To use this command effectively, it is recommended that the dev KEK have the maximum possible length (1194 bytes).  Then, it can be replaced with a production key of any length.

utzig: I don't think that addresses my doubt... let me give you an example with the max size you mentioned: assume I write one key that is 1192, that is OK because it's smaller than 1194. Now I want to write a key of 1194 bytes, it should be OK as well because it equals the max size, but it won't work because it's larger than the previous (old) key. Or I might be misinterpreting the `len(nkey)` and `len(okey)` which to me look like the new key and old key.

ccollins476ad: That is a valid point.  The second replace operation will fail because the new key is too long.  I see that a lot of context is missing from this PR.

The specific use case here is:
1. boot loader / mfgimage built with development keys ("dev mfgimage").
2. dev mfgimage is tested and deemed "good".
3. dev mfgimage is converted to production mfgimage by replacing the dev KEK with the prod KEK (and some other manipulations that are not relevant here).

Now we have a production mfgimage that is ready to be shipped to the factory.  We will never replace the KEK in the prod mfgimage, so it doesn't matter if the prod KEK has a short length.

utzig: Ok, looks good

ccollins476ad:
Thanks, @utzig!

PR #260, opened 2019-01-19
`$import` directive

ccollins476ad:
### PROBLEM:

Within a project, there can be a lot of duplication among target `syscfg.yml` files.  The Mynewt package hierarchy mitigates this problem to some extent: the BSP defines some settings, the app overrides a few of those, and the target overrides yet more.  This does not resolve the problem fully, however.  Often, a project has several targets that differ only by one or two syscfg overrides.  For example, a project may have `dev`, `prod`, and `qa` targets, all nearly identical to each other.  Duplicating all the target-level overrides in each target is like any kind of duplication: error prone and a maintenance burden.

### PROPOSAL:

A new Mynewt keyword: `$import`.  This allows a Mynewt YAML file to import the contents of other YAML files.

Example (`targets/dev/syscfg.yml`):

```
$import:
    - 'targets/some_target/syscfg.yml'
    - 'targets/another_target/syscfg.yml'

syscfg.vals:
    OS_DEBUG_MODE: 1
```

In this example, the `dev` target's `syscfg.yml` file imports the syscfg files of two other targets.  The target inherits all the setting definitions and values specified in the imported files.

This feature is similar, but not identical, to C's `#include` directive.  Differences between the two features are:

1. Like `#include`, import paths can be absolute or relative.  However, relative paths are always relative to the project base, not to the file doing the importing.

2. The `$import` map can be placed anywhere in the file, but its placement does not affect its behavior.  Imports are always processed before the rest of the file.

3. Unlike `#include`, `$import` does not perform a simple text substitution.  If it did, it would result in an invalid YAML file (multiple `syscfg.defs` or `syscfg.vals` maps).  Instead, maps with the same name are collected from all imported files and merged into the importing file.

### NOTES

1. This feature introduces the concept of newt keywords.  Keywords are special directives in a YAML file that alter newt's behavior.  Keywords always start with "$".  If newt encounters a keyword that it doesn't understand, it warns the user.

2. This feature may make it confusing to trace down the source of a syscfg override.  Overrides are still a package-level concept, so the `target config` and `target [rev]dep` output indicate which package overrides a setting, not which file.  The implementation of the `$import` feature attaches file info to each syscfg setting (file that defines or overrides the setting along with its import tree), but this information is not accessible at the moment.  I imagine we will want an alternative `target config` command which displays the file performing the override rather than the package.

### DIFF
--- a/util/fileinfo.go
+++ b/util/fileinfo.go
@@ -1,0 +1,43 @@
+package util
+
+import (
+	"fmt"
+)
+
+// FileInfo represents a configuration source.  It is intended to help the user
+// understand how the system acquired its configuration, and to aid in tracking
+// down errors in configuration files.
+type FileInfo struct {
+	Path   string    // Path of configuration file.
+	Parent *FileInfo // File that imported this configuration file, if any.
+}
+
+// ImportString creates a string describing the import hierarchy of the given
+// FileInfo.  It should be called on the *parent* of the file of interest.
+func (fi *FileInfo) ImportString() string {
+	s := ""
+
+	first := true
+	for fi != nil {
+		if !first {
+			s += ", "
+		}
+		first = false
+
+		s += fmt.Sprintf("imported from %s", fi.Path)
+		fi = fi.Parent
+	}
+
+	return s
+}
+
+// ErrTree decorates the given error with a description of the configuration
+// file's import hierarchy.  If a configuration error is encountered, this
+// function should be called on the configuration file's *parent*.
+func (fi *FileInfo) ErrTree(err error) error {
+	if fi == nil {
+		return err
+	}
+
+	return FmtNewtError("%s - %s", err.Error(), fi.ImportString())
+}
--- a/newt/cli/vars.go
+++ b/newt/cli/vars.go
@@ -25,10 +25,11 @@ import (
 	"sort"
 	"strings"
 
+	"mynewt.apache.org/newt/newt/config"
 	"mynewt.apache.org/newt/newt/interfaces"
-	"mynewt.apache.org/newt/newt/newtutil"
 	"mynewt.apache.org/newt/newt/pkg"
 	"mynewt.apache.org/newt/newt/project"
+	"mynewt.apache.org/newt/newt/toolchain"
 	"mynewt.apache.org/newt/util"
 )
 
@@ -79,8 +80,8 @@ func buildProfileValues() ([]string, error) {
 
 	packs := project.GetProject().PackagesOfType(pkg.PACKAGE_TYPE_COMPILER)
 	for _, pack := range packs {
-		v, err := newtutil.ReadConfig(pack.(*pkg.LocalPackage).BasePath(),
-			"compiler")
+		v, err := config.ReadFile(
+			pack.BasePath() + "/" + toolchain.COMPILER_FILENAME)
 		if err != nil {
 			return nil, err
 		}
--- a/newt/config/config.go
+++ b/newt/config/config.go
@@ -1,0 +1,205 @@
+// The config package handles reading of newt YAML files.
+package config
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/spf13/cast"
+
+	"mynewt.apache.org/newt/newt/newtutil"
+	"mynewt.apache.org/newt/newt/ycfg"
+	"mynewt.apache.org/newt/util"
+	"mynewt.apache.org/newt/yaml"
+)
+
+const (
+	KEYWORD_IMPORT = "$import"
+)
+
+// keywordMap is a map of all supported keywords.  Config keywords always start
+// with "$".
+var keywordMap = map[string]struct{}{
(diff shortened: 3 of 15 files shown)


utzig:
I am curious on why you call it `$import` instead of just `import`, aren't those map keys already keyword behaving just by being a map key in the yaml?

ccollins476ad:
Thanks @utzig, that is a good question.  I didn't explain it well, but these `$` keywords apply equally to all YAML files, whereas maps like `pkg.cflags` or `syscfg.defs` are specific to the file type.  The process goes something like this:

```
For each YAML file f:
    Process all keywords
    Construct a tree from the non-keyword content of the file
```

So, keywords are processed while the tree is being constructed.  Only after the tree has been constructed can non-keyword entries be processed.

We could remove the `$` and still get the same functionality, but then we wouldn't be able to warn about unrecognized keywords.  Alternatively, the tree construction process could just "know" all the valid top-level strings, and warn about everything it doesn't recognize.  But I think this blurs the phases of transformation that newt uses to read the project, and would add some messiness to the code.

benmccrea:
@ccollins476ad I'd appreciate it if you can clarify how overrides work. For your example above, what would happen in these two cases:
A) `targets/some_target/syscfg.yml` contains `OICMGR_TRANS_SECURITY: 0`
and  `targets/another_target/syscfg.yml` contains `OICMGR_TRANS_SECURITY: 1`
B) `targets/some_target/syscfg.yml` contains `OS_DEBUG_MODE: 0`

ccollins476ad:
Thanks @benmccrea, those are good questions.  It turns out my testing of this feature left much to be desired; I discovered a bug while coming up with a response!  I have pushed the fix.

I am using different settings here than the specific ones you asked about (`OS_DEBUG_MODE`->`HAL_FLASH_VERIFY_ERASES`, `OICMGR_TRANS_SECURITY`->`HAL_FLASH_VERIFY_BUF_SZ`).  This allows a more basic target with a simpler syscfg.  There are three files:

#### `targets/my_blinky_sim/syscfg.yml`:
```
$import:
    - 'misc/importa.yml'
    - 'misc/importb.yml'

syscfg.vals:
    HAL_FLASH_VERIFY_ERASES: 1
```

#### `misc/importa.yml`:
```
syscfg.vals:
    HAL_FLASH_VERIFY_BUF_SZ: 32
```

#### `misc/importb.yml`:
```
syscfg.vals:
    HAL_FLASH_VERIFY_BUF_SZ: 16
```

Newt generated the following HAL syscfg for this target:
```
[@apache-mynewt-core/hw/hal]
  HAL_FLASH_VERIFY_BUF_SZ: 16 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_ERASES: 1 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_WRITES: 0
```

If we reverse the order of the imports (i.e., `importb` before `importa`), then the syscfg changes:
```
[@apache-mynewt-core/hw/hal]
  HAL_FLASH_VERIFY_BUF_SZ: 32 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_ERASES: 1 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_WRITES: 0
```

So, import order matters, which I think is the correct behavior.  This is something that should be documented.  Imports are processed in the specified order, then the importing file is processed.

util/fileinfo.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// ImportString creates a string describing the import hierarchy of the given
+// FileInfo.  It should be called on the *parent* of the file of interest.
+func (fi *FileInfo) ImportString() string {
+	s := ""
+
+	first := true
+	for fi != nil {
+		if !first {
```
utzig: if `first` starts with `true` will the contents of this `if` ever execute?

utzig:
This looks good, but one suggestion and doubt...

From your first example above (Setup 1)
```
[@apache-mynewt-core/hw/hal]
  HAL_FLASH_VERIFY_BUF_SZ: 16 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_ERASES: 1 (overridden by targets/my_blinky_sim)
  HAL_FLASH_VERIFY_WRITES: 0
```
It could be interesting to mention that `HAL_FLASH_VERIFY_BUF_SZ` override value came from an import. Also if there was no override, as in the imported yml defined a new value, it would not be obvious where it came from, correct?

util/fileinfo.go
ccollins476ad: Ah, you are right - the `first = false` statement should be outside the `if`.

ccollins476ad:
Thanks, @utzig.  I agree - people will want to know which file each setting define / override comes from.  The mechanism is there (file info attached to each node in the syscfg tree), but I wasn't sure of the best way to present this information to the user.  This question seems like it could use a bit more thought before committing to anything.  Do you have any ideas here?

utzig:
@ccollins476ad 

```HAL_FLASH_VERIFY_BUF_SZ: 16 (overridden by targets/my_blinky_sim [misc/importb])```

^ this would convey all necessary information, but I don't think it is actually a very important change, unless people really abuse `$import`! ;-)

PR #266, opened 2019-02-05
`newt target dump` command 

ccollins476ad:
**Note:** If this PR is merged, it should be merged along with https://github.com/JuulLabs-OSS/mynewt-travis-ci/pull/27.

This command produces a JSON dump of a target's intermediate representation.  This representation includes:

* Dependency graph
* Reverse dependency graph
* Syscfg tree
* Sysinit sequence
* Sysdown sequence
* Log configuration tree
* API resolution
* Flash map
* Warnings (unsatisfied APIs, API conflicts)

Newt calculates this information to use as input to a build (and for other newt commands).  The output of the `newt target dump` command can be used for regression testing - if a target's intermediate representation is not equivalent to its known good value, then newt is broken!

Unfortunately, this PR contains quite a number of additional code changes.  These changes are either 1) fixes to bugs discovered during implementation, or 2) required by the command itself.


utzig:
@ccollins476ad This needs rebase / conflict resolution (after merging #268, I think).

ccollins476ad:
> @ccollins476ad This needs rebase / conflict resolution (after merging #268, I think).

Thanks, @utzig.  I have fixed it.

newt/builder/depgraph.go
```diff
@@ -92,58 +91,86 @@ func revdepGraph(rs *resolve.ResolveSet) (DepGraph, error) {
...
-
-		// Add nodes for packages with with dependers.
-		for _, child := range children {
-			rParent := child.Rpkg
-			rChild := *child
-			rChild.Rpkg = parent
-			graphMapAdd(revGm, rParent, &rChild)
+	rgraph := DepGraph{}
+	for parent, entries := range graph {
+		// Ensure parent not is present in graph, even if no one depends on it.
```
utzig: Yoda comment style!

newt/builder/depgraph.go
```diff
@@ -92,58 +91,86 @@ func revdepGraph(rs *resolve.ResolveSet) (DepGraph, error) {
...
+			reqes := entry.ReqApiExprs[api]
+			reqdis := reqes.Disjunction().String()
+
+			apies := entry.ApiExprs[api]
+			apidis := apies.Disjunction().String()
 
-	if len(reasons) > 0 {
-		s += fmt.Sprintf("(%s)", strings.Join(reasons, " "))
+			s += "(api:" + api
+			if reqdis != "" || apidis != "" {
```
utzig: Won't the resulting string be a bit strange if `reqdis == ""` and `apidis != ""`?

newt/builder/depgraph.go
```diff
@@ -92,58 +91,86 @@ func revdepGraph(rs *resolve.ResolveSet) (DepGraph, error) {
...
 	}
 
 	return s
 }
 
 func DepGraphText(graph DepGraph) string {
-	parents := make([]*resolve.ResolvePackage, 0, len(graph))
-	for lpkg, _ := range graph {
-		parents = append(parents, lpkg)
+	parents := make([]string, 0, len(graph))
```
utzig: This same declaration style could also be used on the code above, `var apis []string`, to get a bit of extra performance (maybe)

newt/parse/expr.go
```diff
@@ -0,0 +1,75 @@
+package parse
+
+// ExprSet is a set of expressions.  The map value is the expression itself;
+// the key is the string representation of the expression, obtained via
+// `Node#String()`.  This is implemented as a map in this way to prevent
+// duplicate expreessions.
```
utzig: typo `expressions`

newt/syscfg/restrict.go
```diff
@@ -82,6 +84,48 @@ type CfgRestriction struct {
...
+func (c CfgRestrictionCode) MarshalJSON() ([]byte, error) {
+	s := c.String()
+	j, err := json.Marshal(s)
+	if err != nil {
+		return nil, util.ChildNewtError(err)
+	}
+	return j, nil
+}
+
+func (c *CfgRestrictionCode) UnmarshalJSON(b []byte) error {
```
utzig: These marshal / unmarshall functions look pretty much the same that the ones in `marshal.go` with a different struct being marshalled and a different string parsing functions, perhaps they could be standalone and called passing the string parsing function?

newt/builder/depgraph.go
ccollins476ad: Yes, it does look a bit strange in this case.  As an example, when `sys/log/full`'s `log` API is made conditional on `OS_DEBUG_MODE`:
```
pkg.apis.OS_DEBUG_MODE:
    - log
```

then `newt revdep ...` produces the following:
```
@apache-mynewt-core/sys/log/full <-- [@apache-mynewt-core/sys/log/modlog(api:log,syscfg:;OS_DEBUG_MODE)
```

The semicolon looks weird, but at least the output is unambigous (the structure is always `<req-api-expr>;<api-expr>`.  I sort of gave up on presenting all this information in a "pretty" way, and settled for something that can be unambiguously translated back to source.  Suggestions for making this output more readable are much appreciated! :)

newt/syscfg/restrict.go
ccollins476ad: That's a good idea!  I think we could eliminate a lot of duplication with something like this:
```
func MarshalJSONStringer(sr fmt.Stringer) ([]byte, error) {
    s := sr.String()
    j, err := json.Marshal(s)
    if err != nil {
        return nil, ChildNewtError(err)
    }

    return j, nil
}
```

(`fmt.Stringer` is an interface that supports the `String()` function).

Then the `CfgRestrictionCode` `MarshalJSON` function becomes:
```
func (c CfgRestrictionCode) MarshalJSON() ([]byte, error) {
    return util.MarshalJSONStringer(c)
}
```

Unfortunately, I don't think we can do something similar for the unmarshal functions.  The problem is that each unmarshal function needs to return a different type, and Go doesn't support generics.  Still, eliminating the duplication in all the marshal functions is great.

newt/builder/depgraph.go
ccollins476ad: Oops, left an extra "not" in there :).

newt/builder/depgraph.go
ccollins476ad: Good catch, thanks.  I have fixed it.

ccollins476ad:
Thanks for reviewing this, @utzig.  I fixed the issues you identified and extracted the common JSON marshaling code as I mentioned in my comment.

PR #272, opened 2019-02-27
Fix expressions in settings restrictions

andrzej-kaczmarek:
This patch makes following restrictions possible:
```
  UART_1:
      description: 'Enable nRF52xxx UART1'
      value: 0
      restrictions:
          - 'MCU_TARGET == "nRF52840"'
```
Currently this would be evaluated without translation thus `UART_1` value
does not matter here (as in package restriction). Instead, we want to
make this translated into `(MCU_TARGET == "nRF52840") || (!UART_1)`
which is what this patch does.

### DIFF
--- a/newt/syscfg/restrict.go
+++ b/newt/syscfg/restrict.go
@@ -158,7 +158,7 @@ func translateShorthandExpr(expr string, baseSetting string) string {
 	}
 
 	if ifi == -1 {
-		if parse.FindBinaryToken(tokens) == -1 {
+		if parse.FindBinaryToken(tokens) == -1 || len(baseSetting) > 0 {
 			// [!]<req-setting>
 			return fmt.Sprintf("(%s) || !%s", expr, baseSetting)
 		} else {


ccollins476ad:
I understand the value of this change, but I think it introduces some unintended effects.  Specifically, the value `0` is now valid for all settings (unless there is a package-level restriction that prevents it).

For example, if I add the following restriction to the `SHELL_MAX_COMPAT_COMMANDS` setting:
```
    SHELL_MAX_COMPAT_COMMANDS:
        description: 'Max number of compatibility commands'
        value: 20
        restrictions:
            - SHELL_MAX_COMPAT_COMMANDS > 10
```

In the current newt, overriding this setting with any value less than 10 triggers a restriction error.

With the patch, values in the range 1-9 trigger the error, but a value of 0 is accepted.  As I mentioned, this can be worked around with a package-level restriction.  Is this behavior intended?

I think the root problem is that the original restriction notation was implemented without enough thought put into it.  Now that we support more complex expressions, maintaining the old user-friendly notation is becoming more difficult.

andrzej-kaczmarek:
Good catch. No, this is definitely not an intended behavior.
For now I can expand those expressions "manually", but I'll think about a better solution here. Just off the top of my head, perhaps this would work if we only apply the fix for expressions that do not reference `baseSetting`?

ccollins476ad:
Closing this.  It looks like it stalled about a year ago.  Please reopen as needed.

PR #273, opened 2019-02-27
Allow to specify setting value range

andrzej-kaczmarek:
This allows to specify allowed range for numeric syscfg values in a simple way instead of requiring complex expressions. It allows for any combinations of discrete values and value ranges to be used in single restriction.

`range` value is comma-separated list of ranges and discrete values which is then converted to proper expression and evaluated, for example:
```
    range: 1,2,4,8
    range: 0,10..15
    range: 5..10,15..20,99
```
Also it is possible to reference other settings:
````
    BLE_LL_MAX_PKT_SIZE:
        value: '251'
        range: 27..251
    BLE_LL_SUPP_MAX_RX_BYTES:
        value: 'MYNEWT_VAL(BLE_LL_MAX_PKT_SIZE)'
        range: 27..BLE_LL_MAX_PKT_SIZE
    BLE_LL_SUPP_MAX_TX_BYTES:
        value: 'MYNEWT_VAL(BLE_LL_MAX_PKT_SIZE)'
        range: 27..BLE_LL_MAX_PKT_SIZE
    BLE_LL_CONN_INIT_MAX_TX_BYTES:
        value: '27'
        range: 27..BLE_LL_SUPP_MAX_TX_BYTES
```

### DIFF
--- a/newt/syscfg/restrict.go
+++ b/newt/syscfg/restrict.go
@@ -71,20 +71,30 @@ const (
 	CFG_RESTRICTION_CODE_NOTNULL = iota
 	CFG_RESTRICTION_CODE_EXPR
 	CFG_RESTRICTION_CODE_CHOICE
+	CFG_RESTRICTION_CODE_RANGE
 )
 
 var cfgRestrictionNameCodeMap = map[string]CfgRestrictionCode{
 	"$notnull": CFG_RESTRICTION_CODE_NOTNULL,
 	"expr":     CFG_RESTRICTION_CODE_EXPR,
 	"choice":   CFG_RESTRICTION_CODE_CHOICE,
+	"range":    CFG_RESTRICTION_CODE_RANGE,
+}
+
+type CfgRestrictionRange struct {
+	LExpr string
+	RExpr string
 }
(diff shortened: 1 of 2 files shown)


ccollins476ad:
I forgot to mention - there is a missing import in `syscfg/restriction.go` (`strings`).  But aside from that, the code looks good!

andrzej-kaczmarek:
@ccollins476ad 
re2 that would be nice, but I do not quite know how we would determine whether `-` is a range or subtraction operator? we can use `..` instead of `-` and then ranges can be translated easily into more complex expressions so we could have things like `1,2,3,5 .. FOOVAL-5`
re3 good point, will fix
re5 not sure how it is missing here - it is in my local tree :) will fix.

sjanc:
.. for range sounds good

andrzej-kaczmarek:
@ccollins476ad @sjanc updated, `..` is now used to indicate range and warning is emitted when `lval > rval` in a range. this should also allow for ranges like `0..FOO-2` but apparently expressions do not yet support arithmetic operators :-)

PR #274, opened 2019-02-28
Remember which packages satisfy API requirements

ccollins476ad:
This is a fix for an error reporting bug.  When any API requirement was unsatisfied, newt would report all APIs as unsatisfied.

For example, mynewt-blinky at commit
`06536d2a86a4eaf8d619fe2bad75a4cdf9b77cc8` is broken.  The blinky app has an unsatisfied `log` dependency.  When I try building it, I get the following error:

```
    Error: Unsatisfied APIs detected:
        * console, required by: hw/mcu/native, kernel/os
        * log, required by: sys/log/modlog
```

The `console` API is satisfied, so it should not be reported.

This newt bug was introduced in `a2573a97549cc05dd4949ffca3f87165b239f8aa`.  Prior to that commit, each satisfied API dependency was tracked in two directions: the depender remembers which package satisfies its API requirement, and the dependee remembers who requires the API.  In the bad commit, the first of these was accidentally removed.

The fix is to add the missing code back.  Now, the depender remembers which packages satisfy its API requirements again.

NOTE: This bug would have been caught by the "dump" tests in https://github.com/JuulLabs-OSS/mynewt-travis-ci... if we were testing targets with unsatisfied API dependencies!  I plan to add a bunch more targets to that test to catch problems like this.

### DIFF
--- a/newt/parse/expr.go
+++ b/newt/parse/expr.go
@@ -9,6 +9,16 @@ type ExprSet map[string]*Node
 // ExprMap is a collection of named expression sets.
 type ExprMap map[string]ExprSet
 
+func NewExprSet(exprs []*Node) ExprSet {
+	if len(exprs) == 0 {
+		return nil
+	}
+
+	es := make(ExprSet, len(exprs))
+	es.Add(exprs)
+	return es
+}
+
 // Exprs returns a slice of all expressions contained in the expression set.
 // The resulting slice is sorted.
 func (es ExprSet) Exprs() []*Node {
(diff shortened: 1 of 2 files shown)


utzig:
This look OK to me, I re-started the CI to check if it passes this time...

utzig:
@ccollins476ad It seems you'll need to update https://github.com/JuulLabs-OSS/mynewt-travis-ci/blob/master/newt_dump/answers/my_blinky_sim.json#L1566 to remove "console" from the unsatisfied_apis after this one is merged!

PR #275, opened 2019-03-07
Always report the reason for rebuilding a file

ccollins476ad:
Prior to this commit, newt only reported a rebuild reason when it compiled a source file.

Now, newt reports a reason whenever it rebuilds any file (source file, archive, or elf).  The message also includes both filenames - source and destination.

### DIFF
--- a/newt/toolchain/deps.go
+++ b/newt/toolchain/deps.go
@@ -21,6 +21,7 @@ package toolchain
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


newt/toolchain/deps.go
```diff
@@ -244,6 +272,7 @@ func (tracker *DepTracker) CompileRequired(srcFile string,
 		}
 
 		if depModTime.After(objModTime) {
+			logRebuildReqdNewDep(srcFile, dep)
 			util.StatusMessage(util.VERBOSITY_VERBOSE, "%s - rebuild required; obj older than dependency (%s)\n", srcFile, dep)
```
utzig: This line can be removed, no?

ccollins476ad: Right you are.  Thanks, @utzig.  I have removed the extra log line,

PR #278, opened 2019-03-16
Add newt docs command to automatically build using sphinx

sterlinghughes:
This automatically builds documents within a mynewt repo, including any sub-repos using the sphinx format.

### DIFF
--- a/newt/docs/docsbuilder.go
+++ b/newt/docs/docsbuilder.go
@@ -1,0 +1,145 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


utzig:
Whata is your main design idea, to build docs for local consumption or to push to the main docs website?

util/util.go
```diff
@@ -502,6 +502,21 @@ func MoveDir(srcDir string, destDir string) error {
 	return nil
 }
 
+func CallInDir(path string, execFunc func() error) error {
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	os.Chdir(path)
```
ccollins476ad: It is probably a good idea to check for errors here.

Also, this is just a stylistic comment - I think it is more "Go like" to switch back to the original directory using `defer`.  Then you can continue to return on error.

newt/docs/docsbuilder.go
```diff
@@ -0,0 +1,145 @@
...
+func (db *DocsBuilder) GetDocs() ([]*DocDescriptor, error) {
+	docs := []*DocDescriptor{}
+
+	for _, repo := range db.project.Repos() {
+		name := repo.Name()
+
+		if repo.IsLocal() {
+			name = db.project.Name()
+		}
+		fmt.Println(name)
```
utzig: Should be a debug message

utzig:
@ccollins476ad I think this is a good start, are you OK with merging and updating later?

ccollins476ad:
@utzig, sounds good.

PR #284, opened 2019-04-10
Fix build when not a git tree

utzig:


utzig:
This fixes when running a package that is not a git clone, like when the user downloads from here: https://github.com/apache/mynewt-newt/releases

Bash script should beOK, but I decided to skip the `NewtGitHash: unknown` text in those cases, which may be OK or not...

newt/newt.go
```diff
@@ -140,7 +140,9 @@ func newtCmd() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			fmt.Printf("Apache Newt\n")
 			fmt.Printf("   Version: %s\n", newtutil.NewtVersionStr)
-			fmt.Printf("  Git Hash: %s\n", newtutil.NewtGitHash)
+			if newtutil.NewtGitHash != "unknown" {
+				fmt.Printf("  Git Hash: %s\n", newtutil.NewtGitHash)
```
ccollins476ad: I don't have a strong feeling about this, but I think it is better to be explicit here.  I would just print "unknown" if we don't know the git hash.

utzig: done!

PR #287, opened 2019-04-17
Use 2.0 image format by default

sjanc:
Now that Mynewt switched to mcuboot and removed legacy boootutils it
make sense to use new format by default.


sjanc:
Should be merged after https://github.com/apache/mynewt-core/pull/1763

utzig:
I think `mfgimage` refers to manufacturing image; what this is doing is a change to generate images in a format compatible with MCUBoot (changes image header format).

PR #289, opened 2019-04-17
Allow target key_file to resolve pkgpath

utzig:


utzig:
@ccollins476ad This relates to https://github.com/JuulLabs-OSS/mcuboot/pull/470. Currently the CI fails because it cannot build the targets found in mcuboot directory (now a dependency!) because if key_file paths. This adds support for using pkgpaths in the key_files.

PR #299, opened 2019-05-21
Add ed25519 signing support

utzig:
Needs testing in HW; needs update to `go mod` (from dep)


utzig:
Related to https://github.com/JuulLabs-OSS/mcuboot/pull/479

ccollins476ad:
Looks good to me.  I think the `go.mod` and `go.sum` files need to be updated.  This should do the trick:

```
rm go.mod go.sum
GO111MODULE=on go mod init
cd newt
GO111MODULE=on go build
```

This is not the usual way to update these files, but since you already updated the `Gopkg.[...]` files, it is easier to rebuild the `go.[...]` files from them.

We should remove `Gopkg.lock` and `Gopkg.toml` to avoid confusion, but that can happen in a different PR.

sterlinghughes:
yes, and can we _please_ remove vendor/ soon 👍

utzig:
@ccollins476ad Ok, so now vendoring is 100% removed, I am a little bit in doubt if I should leave the go requirement to 1.12 or try rolling back to 1.11?

ccollins476ad:
The code looks good to me, but CI is failing:
```
go build golang.org/x/sys/unix: module requires Go 1.12
go build github.com/spf13/cobra: module requires Go 1.12
```

Out of curiosity, why did you remove this from `.travis.yml`:
```
env:
  - GO111MODULE=on
```
?

utzig:
@ccollins476ad The CI is failing because I downgraded back to 1.11 to see if it works. It works on my computer with go 1.11.5 but not 1.11.2. And the `GO111MODULE` env removal was also for testing, wanted to see if it would pass, because it passes on my computer (it did on travis-ci).

This is the build that passed: https://travis-ci.org/apache/mynewt-newt/builds/549843583

utzig:
@ccollins476ad Btw, it looks like the reason it was passing before is that `spf13/cobra` and `x/sys/unix` were already present under `vendor/` so they were being used instead of grabbing from online sources. So one alternative is to roll-back the removal of those two packages, or instead settle on 1.12 as a requirement.

ccollins476ad:
I think it is easiest to just require a newer version of Go (1.12 or "1.x").

PR #303, opened 2019-06-21
travis: enable Go module support

ccollins476ad:
Now that newt uses modules (https://github.com/golang/go/wiki/Modules), the `vendor` directory is no longer needed.


sterlinghughes:
yay!  this needed to go for awhile.

PR #309, opened 2019-07-16
Allow to override path escaping from command line

sjanc:
With this patch it is possible to disable path escaping on Windows.
It looks like some shells on Windows require no escaping.

### DIFF
--- a/newt/newt.go
+++ b/newt/newt.go
@@ -91,9 +91,7 @@ func newtCmd() *cobra.Command {
 				verbosity = util.VERBOSITY_VERBOSE
 			}
 
-			if newtEscapeShellCmds {
-				util.EscapeShellCmds = true
-			}
+			util.EscapeShellCmds = newtEscapeShellCmds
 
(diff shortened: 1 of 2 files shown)


newt/newt.go
```diff
@@ -128,7 +126,7 @@ func newtCmd() *cobra.Command {
 	newtCmd.PersistentFlags().BoolVarP(&newtHelp, "help", "h",
 		false, "Help for newt commands")
 	newtCmd.PersistentFlags().BoolVarP(&newtEscapeShellCmds, "escape", "",
-		false, "Apply Windows escapes to shell commands")
+		runtime.GOOS == "windows", "Apply Windows escapes to shell commands")
```
utzig: Does it really matter to turn it on by default for Windows here, if above you have `util.EscapeShellCmds = newtEscapeShellCmds`?

sjanc: I wanted to keep default behavior same as in 1.6

PR #313, opened 2019-07-22
During install / upgrade, if the requested commit doesn't exist, try to check out the latest "rc" tag instead

ccollins476ad:
During install / upgrade, if the requested commit doesn't exist, try to check out the latest "rc" tag instead.  This behavior is useful when a release candidate is being tested.  In this case, the "rc" tags exist, but the official release tag has not been created yet.

## Example:

```
### project.yml
        type: git
        vers: 2.0.0

### repository.yml (mynewt-core)
    repo.versions:
        "2.0.0": "mynewt_2_0_0_tag"
```

The `mynewt_2_0_0_tag` doesn't actually exist yet.  This tag will not get created until 2.0.0 is actually released.  However, the following tags do exist:

* `mynewt_2_0_0_rc1_tag`
* `mynewt_2_0_0_rc2_tag`

When newt attempts to upgrade, upon discovering that the requested tag does not exist, it uses the latest "rc" tag instead (rc2 in this case).  If an "rc3" tag is later added, then the next upgrade operation will check out the new tag.

### DIFF
--- a/newt/downloader/downloader.go
+++ b/newt/downloader/downloader.go
@@ -25,7 +25,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


utzig:
@ccollins476ad Please rebase!

ccollins476ad:
Thanks @utzig.  I have fixed it.

PR #316, opened 2019-07-31
cmake.go: Replace backslashes in some paths

michal-narajowski:
CMake on Windows doesn't handle backslashes well so we replaces
backslashes with slashes.

This fixes https://github.com/apache/mynewt-newt/issues/315


kasjer:
Verified to work on Windows

PR #318, opened 2019-08-01
Remove `newt install`

ccollins476ad:
`newt upgrade` does everything that `newt install` does.  Now, `newt install` warns the user and performs an upgrade.

### DIFF
--- a/newt/cli/project_cmds.go
+++ b/newt/cli/project_cmds.go
@@ -107,8 +107,8 @@ func installRunCmd(cmd *cobra.Command, args []string) {
 	interfaces.SetProject(proj)
 
 	pred := makeRepoPredicate(args)
-	if err := proj.InstallIf(
-		false, newtutil.NewtForce, newtutil.NewtAsk, pred); err != nil {
+	if err := proj.UpgradeIf(
+		newtutil.NewtForce, newtutil.NewtAsk, pred); err != nil {
 
 		NewtUsage(nil, err)
 	}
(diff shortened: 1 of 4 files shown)


newt/cli/project_cmds.go
```diff
@@ -200,7 +202,7 @@ func AddProjectCommands(cmd *cobra.Command) {
 	installHelpEx += "    Installs the apache-mynewt-core repository."
 	installCmd := &cobra.Command{
 		Use:     "install [repo-1] [repo-2] [...]",
-		Short:   "Install project dependencies",
+		Short:   "Install project dependencies (deprecated)",
```
utzig: This looks fine to me, but there is also a `Deprecated` key. You could change this line to `Deprecated: "Install project dependencies"` and it would print a standard deprecation message (and still work).

ccollins476ad: @utzig  Thanks for the tip (and the review).  That is pretty cool.  I have changed "install" and "sync" to use `Deprecated`.

sterlinghughes:
I don't think we should deprecate install without updating documentation.  Also, why not just make install a synonym for upgrade -- I think calling "upgrade" is confusing when installing repo.

ccollins476ad:
@sterlinghughes, with this PR, "install" just executes an "upgrade".  It does print a deprecated warning, however.  You're right about updating the docs...

Over the years there have been complaints about too many project commands (install, upgrade, sync).  I thought we all wanted to reduce the number.  I dislike having multiple commands that do the same thing (and personally I don't find "upgrade" confusing when starting a new project).  Here is what I am thinking - I will "undeprecate" the install command and send an email to the dev list about this change to see what people think.  How does that sound?

PR #321, opened 2019-08-09
Add MYNEWT_PROJECT_ROOT env variable

nkaje:
This variable defines a standard way to access the root of the project.

### DIFF
--- a/newt/builder/paths.go
+++ b/newt/builder/paths.go
@@ -31,6 +31,10 @@ import (
 const BUILD_NAME_APP = "app"
 const BUILD_NAME_LOADER = "loader"
 
+func ProjectRoot() string {
+	return project.GetProject().Path()
+}
+
 func BinRoot() string {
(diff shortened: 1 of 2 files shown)


newt/builder/paths.go
```diff
@@ -31,6 +31,10 @@ import (
 const BUILD_NAME_APP = "app"
 const BUILD_NAME_LOADER = "loader"
 
+func ProjectRoot() string {
+	return project.GetProject().Path() + "/"
```
ccollins476ad: Just a nitpick, but I prefer when path strings do not include a trailing slash.  Inevitably, people construct a bigger path by appending "/rest/of/path", and the result is a double slash in the middle of the string.  Also, the other paths in `paths.go` do not include a trailing slash.

nkaje: removed trailing '/' in the latest PS

PR #325, opened 2019-08-21
Shellcmd override sigblock

nkaje:
Adding a separate optional step to check the status of the JTAG lock status of the target. While adding this, reverting the earlier approach to run 'load' as an interactive command to achieve the same.


ccollins476ad:
Thanks, it looks good.  Could you please run `gofmt` on the modified files?  The formatting looks mostly good, but I think there is some extra whitespace.

nkaje:
Latest patch set uploaded after running it through `gofmt`

ccollins476ad:
Thanks @nkaje.  Sorry I did not mention this sooner... I'm not crazy about using `...` to make `ShellInteractiveCommand()` take an optional argument.  This function doesn't seem to be called in very many places.  Making this argument explicit wouldn't be too disruptive, would it?

nkaje:
No, it wouldn't be too disruptive. But I wasn't sure about how we feel about touching all the invocations though they don't directly mandate the change. I did see that Go users aren't fans of using optional arguments either. I will make another revision that makes the flag mandatory and the API called with right values throughout newt.

util/util.go
```diff
@@ -359,6 +359,7 @@ func ShellCommandLimitDbgOutput(
 // @param env                   Additional key=value pairs to inject into the
 //                                  child process's environment.  Specify null
 //                                  to just inherit the parent environment.
+// @param block				    Optional flag to override blocking signals
```
ccollins476ad: This comments seems to be on the wrong function.

newt/builder/load.go
```diff
@@ -52,6 +54,36 @@ func (t *TargetBuilder) Load(extraJtagCmd string) error {
...
+	sigs := make(chan os.Signal, 1)
+	skipblock := true
+	signal.Notify(sigs, syscall.SIGTERM)
+
+	cmd := []string{
+		checkScript,
+	}
+
+	/* Handle Ctrl-C, terminate newt, as it is intended
+	   intended behavior */
```
ccollins476ad: Just a nitpick: "intended" is duplicated here.

PR #333, opened 2019-09-17
Emit toolchain output to stdout

ccollins476ad:
Newt was discarding all toolchain output (compiler, linker, etc).  If there were any compiler warnings, they would not be visible to the user.

This PR makes newt emit all toolchain output.  If there are no warnings, there is no extra output.


grantfreese:
Looks good to me. I've verified this newt tool builds my project correctly, and displays gcc warnings as expected.

PR #336, opened 2019-10-05
Fix some issues with per-package custom commands

ccollins476ad:
This PR fixes a few issues that @andrzej-kaczmarek identified:

> 1. Scripts paths in packages should be relative to package root as otherwise it's impossible to create portable path.
> 2. Would be good if newt can define env for newt and toolchain binaries so scripts are more portable.
> 3. Seems like newt does not overwrite env if it is already set. This causes problem with recursive build and for now I just unset some env that I need in child build.

This PR also contains a change that was requested in the original PR, but which I mistakenly left out:

4. rename: `post_build` --> `post_link`

5. Andrzej reported another issue which this PR fixes: artifacts generated by pre-link scripts do not get linked into the executable until the *next* build.


newt/builder/buildutil.go
```diff
@@ -149,17 +150,35 @@ func (b *Builder) binBasePath() (string, error) {
...
+
+	return m
+}
+
+func ToolchainEnvVars(c *toolchain.Compiler) map[string]string {
+	return map[string]string{
+		"MYNEWT_CC_PATH":  c.GetCcPath(),
+		"MYNEWT_CPP_PATH": c.GetCppPath(),
+		"MYNEWT_AS_PATH":  c.GetAsPath(),
+		"MYNEWT_AR_PATH":  c.GetArPath(),
```
andrzej-kaczmarek: at least `objcopy` is also useful to manipulate objects, but perhaps we should just expose everything so also `objdump` and `objsize`

andrzej-kaczmarek:
works great, just one more small request :-)

ccollins476ad:
@andrzej-kaczmarek Thanks for trying it.  I have added three more environment variables:

* MYNEWT_OBJCOPY_PATH
* MYNEWT_OBJDUMP_PATH
* MYNEWT_SIZE_PATH

PR #338, opened 2019-10-10
Add support for hardware encryption features

agross-korg:
This patch adds a number of features to support the use of hardware
stored encryption keys.  Specifically, options to denote header padding,
image padding, hw key index, and AES-256-CTR encryption were added.

Images created using the hw encryption keys will result in two additional
TLV types being appended to the image.  The first is a TLV that is used to
store the nonce used for the AES-256-CTR.  The second is a TLV that contains
the hw key index being used to decrypt the image.

Signed-off-by: Andy Gross <andy.gross@juul.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/NickBall/go-aes-key-wrap v0.0.0-20170929221519-1c3aa3e4dfc5
-	github.com/apache/mynewt-artifact v0.0.5
+	github.com/apache/mynewt-artifact v0.0.8
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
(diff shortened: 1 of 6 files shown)


utzig:
I don't think we should be maintaining image version 1 when adding new stuff. Do you have any opinions @ccollins476ad?

ccollins476ad:
@utzig I agree.  I think we should consider v1 support as legacy going forward.  I don't think the changes need to be removed from this PR (as long as they get tested).

agross-korg:
So I'd rather not support this for V1.  Should I just error out if these are used with v1?

ccollins476ad:
@agross-korg I think reporting an error and aborting in the case of v1 is totally fine.

ccollins476ad:
@agross-korg It looks like the travis unit tests are failing for this PR.

utzig:
@agross-korg There are still alignment issues, please run `go fmt`.

agross-korg:
ooops. . fixed and repushed

PR #340, opened 2019-10-12
newt upgrade: Fix Repo Commit Checkout

nkaje:
Checking out various repositories to a particular set of commit versions
is broken. Fix this by looking at the dependency graph's node's
required versions.

Example Usage in repository.yml

```
repo.versions:
    "0.0.0" : "master"
    "1.0.0" : "release_1_0_0"

repo.deps:
    apache-mynewt-core:
        type: git
        url: git@github.com:apache/mynewt-core.git
        vers:
            master: 0-dev
            release_1_0_0: 'f3d0ff2ab8db0da99f6b35938caf25c36950e181-commit'

    apache-mynewt-nimble:
        type: git
        url: git@github.com:apache/mynewt-nimble.git
        vers:
            master: 0-dev
            release_1_0_0: '7624744d90c159b8c9ab0bf3d890b0c3a3c43928-commit'
```

Note that the version specifier (such as `release_1_0_0`) can be tag.
The project that houses this repository.yml gets checked out to
`release_1_0_0` and the dependencies get checked out to the versions
specified in the above spec.

Signed-off-by: Naveen Kaje <naveen.kaje@juul.com>

### DIFF
--- a/newt/install/install.go
+++ b/newt/install/install.go
@@ -646,18 +646,29 @@ func (inst *Installer) calcVersionMap(candidates []*repo.Repo) (
 		return nil, deprepo.ConflictError(conflicts)
 	}
 
-	// If project.yml specified any specific git commits, ensure we get them.
+	// Check all repo dependencies for specified git commits, ensure we get them.
+	rg := dg.Reverse()
 	for name, ver := range vm {
-		reqs := inst.reqs[name]
-		if len(reqs) > 0 {
-			keep, err := inst.shouldKeepCommit(name, reqs[0].Ver.Commit)
-			if err != nil {
-				return nil, err
+		for _, node := range rg[name] {
+			if len(node.VerReqs) > 0 {
+				for _, vreq := range node.VerReqs {
+					if vreq.Ver.Commit != "" {
+						keep, err := inst.shouldKeepCommit(name, vreq.Ver.Commit)
+						if err != nil {
+							return nil, err
+						}
+						if keep {
+							if ver.Commit == "" {
+								ver.Commit = vreq.Ver.Commit
+							} else {
+								return nil, util.FmtNewtError("repo %s: multiple commits %s and %s",
+									name, ver.Commit, vreq.Ver.Commit)
+							}
+						}
+						vm[name] = ver
+					}
+				}
 			}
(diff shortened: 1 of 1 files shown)


newt/install/install.go
```diff
@@ -647,17 +647,29 @@ func (inst *Installer) calcVersionMap(candidates []*repo.Repo) (
...
+					if vreq.Ver.Commit != "" {
+						keep, err := inst.shouldKeepCommit(name, vreq.Ver.Commit)
+						if err != nil {
+							return nil, err
+						}
+						if keep {
+							if ver.Commit == "" {
+								ver.Commit = vreq.Ver.Commit
+							} else {
+								return nil, util.FmtNewtError("multiple commits %s and %s",
```
ccollins476ad: It would probably be good to include the repo name in the error message (`name`).

newt/install/install.go
```diff
@@ -647,17 +647,29 @@ func (inst *Installer) calcVersionMap(candidates []*repo.Repo) (
 	}
 
 	// If project.yml specified any specific git commits, ensure we get them.
```
ccollins476ad: This comment could be updated to indicate that we are checking all repo dependencies (not just those from `project.yml`).

newt/install/install.go
```diff
@@ -647,17 +647,29 @@ func (inst *Installer) calcVersionMap(candidates []*repo.Repo) (
...
-		if len(reqs) > 0 {
-			keep, err := inst.shouldKeepCommit(name, reqs[0].Ver.Commit)
-			if err != nil {
-				return nil, err
-			}
-			if keep {
-				ver.Commit = reqs[0].Ver.Commit
+		for _, node := range rg[name] {
+			if len(node.VerReqs) > 0 {
+				// look at all elements of VerReqs array
```
ccollins476ad: This is 100% a nit pick so feel free to ignore it.  I find comments like this one don't provide much signal vs. noise.  I.e., we can tell from the next line that we are looping over the `VerReqs` slice, so no need to say it twice.

nkaje:
Thanks @ccollins476ad with debugging this issue, patch idea and the review. Comments addressed.

PR #344, opened 2019-10-21
mfg: Support the "extra_manifest" field per-target and per-raw

ccollins476ad:
This new field allows an mfgimage definition to specify an "extra_manifest" map for each embedded target or embedded raw.  This map is passed through unchanged to the resulting mfg manifest JSON file with the name "extra".  This field is useful if the user needs to convey additional information to some system that knows how to parse a manufacturing manifest file.

For example, the following `mfg.yml` snippet:

```
    mfg.targets:
        -
            name: 'targets/blinky_dev'
            area: FLASH_AREA_IMAGE_0
            offset: 0
            extra_manifest:
                map:
                    int1: 5
                    int2: 100
                    str: "abc"
                sequence:
                    - 8
                    - 9
                    - 10
```

yields the following `manifest.json` snippet:

```
    "targets": [
      {
        "name": "targets/blinky_dev",
        "offset": 49152,
        "image_path": "targets/0/image.img",
        "hex_path": "targets/0/image.hex",
        "manifest_path": "targets/0/manifest.json",
        "extra": {
          "map": {
            "int1": 5,
            "int2": 100,
            "str": "abc"
          },
          "sequence": [
            8,
            9,
            10
          ]
        }
      }
    ],
```

The above example only consists of a target, but the same could be done with the `mfg.raw` item.

### DIFF
--- a/newt/mfg/decode.go
+++ b/newt/mfg/decode.go
@@ -40,9 +40,10 @@ type DecodedTarget struct {
 }
 
 type DecodedRaw struct {
-	Filename string
-	Area     string
-	Offset   int
+	Filename      string
+	Area          string
+	Offset        int
+	ExtraManifest map[string]interface{}
 }
 
 type DecodedMmrRef struct {
@@ -226,6 +227,13 @@ func decodeRaw(yamlRaw interface{}, entryIdx int) (DecodedRaw, error) {
 			"mfg raw entry missing required field \"filename\"")
 	}
 	dr.Filename = cast.ToString(filenameVal)
+
+	extra, err := decodeExtra(kv, "extra_manifest")
+	if err != nil {
+		return dr, util.FmtNewtError(
+			"in raw entry %d: %s", entryIdx, err.Error())
+	}
+	dr.ExtraManifest = extra
 
 	return dr, nil
 }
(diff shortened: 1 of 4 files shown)


newt/mfg/decode.go
```diff
@@ -110,6 +112,40 @@ func decodeBoolDflt(kv map[string]interface{}, key string,
...
+	// `map[interface{}]interface{}`.  These need to be converted to
+	// `map[string]interface{}` to be JSON compatible.
+	var fixMap func(m map[string]interface{})
+	fixMap = func(m map[string]interface{}) {
+		for k, v := range m {
+			if _, ok := v.(map[interface{}]interface{}); ok {
+				sm := cast.ToStringMap(v)
+				fixMap(sm)
+
+				m[k] = sm
```
utzig: When this adds a new `string` key from the cast, does the previous `interface{}` key remain in the map?

ccollins476ad: This code is a bit confusing.  There are actually two maps involved:
```
m map[string]interface{}
v.(map[interface{}]interface{})
```

(maybe I could have chosen better variable names!)

The only map that gets modified here is `m` which has `string` keys.  The code just replaces the value associated with a particular string key.

PR #350, opened 2019-11-07
Don't emit spurious YCfg parse warnings

ccollins476ad:
### Background

When newt processes a YAML file, it converts it into a tree form called a YCfg.  For example, this YAML excerpt:

```
OS_MAIN_STACK_SIZE: 100
OS_MAIN_STACK_SIZE.'BLE_MAX_CONNECTIONS > 3': 200
OS_MAIN_STACK_SIZE.SHELL_TASK: 300
```

Is represented as the following tree:

```
                     [OS_MAIN_STACK_SIZE (100)]
                    /                          \
      [BLE_MAX_CONNECTIONS > 3 (200)]      [SHELL_TASK (300)]
```

This YCfg object contains three nodes.  Some of these nodes may be "inactive" in the sense that a required setting is not enabled (e.g., if `SHELL_TASK` is disabled then the rightmost node is inactive).  We don't know which nodes are inactive because we don't have a list of enabled settings when the YCfg is created.

We only know which nodes are active when someone retrieves a value from the tree with a "Get" function.  All the Get functions take a `settings` map parameter which allows the tree to be properly interpreted.  With the list of settings available, the Get functions can identify problems in the tree.  For example, if `BLE_MAX_CONNECTIONS` is configured with a value of "foo", then the expression

```
BLE_MAX_CONNECTIONS > 3
```

cannot be evaluated.  When the Get functions spot a problem like this, they print a warning to the console and ignore the problematic node.

(Perhaps we should convert the YCfg to another "more processed" form after we have calculated the list of settings.  That way we wouldn't need to do any processing in the Get functions.  But that is probably a bigger discussion for another time.)

### Problem

It's a little odd to be doing so much evaluation in a Get function, but this design has worked for the most part.  However, a problem arises when a condition involving a numeric comparison is applied to `pkg.deps`.  For example (from `sys/log/full/pkg.yml`):

```
pkg.deps.'LOG_CONSOLE && LOG_VERSION > 2':
```

This is a problem because newt retrieves `pkg.deps` from each package before it has done syscfg resolution.  The issue is that we can't know what syscfg is until we know the full list of packages included in the build (since each package defines and overrides settings).  We also can't know the full list of packages until we know what syscfg is, since some dependencies are conditional on syscfg settings (as in the example above).  So newt has to repeat a few steps in the process until everything has been fully resolved.  But because newt doesn't know the syscfg definition the first time it encounters the above `pkg.deps` specifier, the `LOG_VERSION` setting is not defined.  The comparison of an undefined setting agains `2` fails and newt prints a warning.  This warning is wrong; `LOG_VERSION` is a valid setting with a numeric value, newt just doesn't know that yet.

### Solution

The solution is to:

1. Remove the warning-printing code from the YCfg Get functions.
2. Change the YCfg Get functions so that they return the parse warnings as an `error` object.

Most code that retrieves values from YCfgs just takes the returned error and prints a warning.  In these cases there is no change in behavior.

During dependency resolution however, any returned errors do not get printed to the screen.  Instead, these warnings are retained for later use.  The previous cycle's warnings are cleared each time a new cycle starts.  After resolution has completed, the retained warnings are printed.  In this way, none of the printed warnings are spurious because they were reported during the final iteration.

### Simple Test

1. Build a target that depends on `sys/log/full`. Ensure no warning is displayed.
2. Modify `sys/log/full/pkg.yml` as follows:
```
@@ -46 +46 @@ pkg.deps.LOG_CLI:
-pkg.deps.'LOG_CONSOLE && LOG_VERSION > 2':
+pkg.deps.'LOG_CONSOLE && LOG_VERSIONx > 2':
```
(i.e., change `LOG_VERSION` to `LOG_VERSIONx`

3. Rebuild the target.  Ensure you see a warning like this one:
```
WARNING: /Users/ccollins/repos/juul-platform/repos/apache-mynewt-core/sys/log/full/pkg.yml: cannot apply > to `LOG_VERSIONx`; operand not a number
```

### DIFF
--- a/newt/ycfg/ycfg.go
+++ b/newt/ycfg/ycfg.go
@@ -266,10 +266,17 @@ func (yc *YCfg) find(key string) *YCfgNode {
 	return cur
 }
 
-func (yc *YCfg) Get(key string, settings map[string]string) []YCfgEntry {
+// Get retrieves all nodes with the specified key.  If it encounters a parse
+// error in the tree, it ignores the bad node and continues the search.  All
+// bad nodes are indicated in the returned error.  In this sense, the returned
+// object is valid even if there is an error, and the error can be thought of
+// as a set of warnings.
+func (yc *YCfg) Get(key string,
+	settings map[string]string) ([]YCfgEntry, error) {
+
 	node := yc.find(key)
 	if node == nil {
-		return nil
+		return nil, nil
 	}
 
 	entries := []YCfgEntry{}
@@ -279,15 +286,18 @@ func (yc *YCfg) Get(key string, settings map[string]string) []YCfgEntry {
 		entries = append(entries, entry)
 	}
 
+	var errLines []string
 	for _, child := range node.Children {
 		expr, err := parse.LexAndParse(child.Name)
 		if err != nil {
-			util.OneTimeWarning("%s: %s", yc.name, err.Error())
+			errLines = append(errLines,
+				fmt.Sprintf("%s: %s", yc.name, err.Error()))
 			continue
 		}
 		val, err := parse.Eval(expr, settings)
 		if err != nil {
-			util.OneTimeWarning("%s: %s", yc.name, err.Error())
+			errLines = append(errLines,
+				fmt.Sprintf("%s: %s", yc.name, err.Error()))
 			continue
 		}
 		if val {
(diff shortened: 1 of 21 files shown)


newt/ycfg/ycfg.go
```diff
@@ -393,29 +422,37 @@ func (yc *YCfg) GetValStringSlice(
 		}
 	}
 
-	return vals
+	return vals, getErr
 }
 
+// GetValStringSlice retrieves all entries with the specified key and coerces
```
utzig: `GetValStringSliceNonempty` retrieves...

ccollins476ad:
Thanks for reviewing this @utzig!  I completely missed that it had been approved.

PR #352, opened 2019-11-08
Include custom commands in `newt dump` output

ccollins476ad:
This allows the travis "dump tests" to verify that newt correctly parses the custom command fields in `pkg.yml` files.

NOTE: This PR will break CI.  After it gets merged, I will update the travis tests to expect the changed "dump" output.

### DIFF
--- a/newt/dump/dump.go
+++ b/newt/dump/dump.go
@@ -34,6 +34,9 @@ type Report struct {
 	Syscfg          Syscfg              `json:"syscfg"`
 	Sysinit         Sysinit             `json:"sysinit"`
 	Sysdown         Sysdown             `json:"sysdown"`
+	PreBuildCmds    ExtCmd              `json:"pre_build_cmds"`
+	PreLinkCmds     ExtCmd              `json:"pre_link_cmds"`
+	PostLinkCmds    ExtCmd              `json:"post_link_cmds"`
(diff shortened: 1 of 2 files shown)


utzig:
> NOTE: This PR will break CI. After it gets merged, I will update the travis tests to expect the changed "dump" output.

Wouldn't it be better to just update in the PR already?

ccollins476ad:
> > NOTE: This PR will break CI. After it gets merged, I will update the travis tests to expect the changed "dump" output.

> Wouldn't it be better to just update in the PR already?

It is the mynewt-travis-ci repo which needs to be updated.  Specifically, the `newt_dump/generate-answers.sh` script needs to be run within that repo.  I can't think of a way to get this change in without breaking CI for a bit.

I can at least submit a PR against the travis repo so that we can merge both PRs at the same time (maybe that is what you meant?).  That would be better.

utzig:
> I can at least submit a PR against the travis repo so that we can merge both PRs at the same time (maybe that is what you meant?). That would be better.
>
> EDIT: I have created JuulLabs-OSS/mynewt-travis-ci#61

Yes, there is no nice way; one option would be to edit `.travis.yml` to checkout the `mynewt-travis-ci` branch that matches this change, but then you would still need to drop it before merge... I guess it's just easier to fix the dump checks later cause you created them in the first place, so you know what you want to check against...

PR #361, opened 2020-01-14
Allow repo hash to correspond to version number

ccollins476ad:
When resolving repo versions, newt uses this procedure:

1. Resolve inter-repo dependencies.
2. Resolve project.yml dependencies. Allow project.yml to override dependencies from step 1.

There was a bug affecting both steps in the procedure.  If a dependency specifies a commit hash that corresponds exactly to a version number, then the dependency was ignored.

This PR fixes the problem so that commit hashes are always honored, whether or not they correspond to version numbers.


nkaje:
Tested and verified with platform repo system.

PR #363, opened 2020-01-15
newt upgrade: Fix repo conflict error message

ccollins476ad:
If a `newt upgrade` operation fails due to conflicting repo requirements, the error message should indicate which requirements are in conflict.  The bug was that the error message was missing "root requirements" (requirements specified in project.yml), so the message
was incomplete.

For example (before fix):
```
    Error: Repository conflicts:
        Installation of repo "mynewt-dummy-repo-04" is blocked:
            mynewt-dummy-repo-01,1.0.0 requires mynewt-dummy-repo-04 ==1.0.0
```
After fix:
```
    Error: Repository conflicts:
        Installation of repo "mynewt-dummy-repo-04" is blocked:
                           project.yml requires mynewt-dummy-repo-04 ==1.0.2/0eaf06337c8f8dc8c49fffa884a6a0c2fb223d73
            mynewt-dummy-repo-01,1.0.0 requires mynewt-dummy-repo-04 ==1.0.0
```

### DIFF
--- a/newt/deprepo/deprepo.go
+++ b/newt/deprepo/deprepo.go
@@ -238,15 +238,20 @@ func PruneMatrix(m *Matrix, repos RepoMap, rootReqs RequirementMap) error {
 // of repos in the project.
 func PruneDepGraph(dg DepGraph, keep []*repo.Repo) {
 	for k, _ := range dg {
-		found := false
-		for _, r := range keep {
-			if r.Name() == k.Name {
-				found = true
-				break
+		// The empty string indicates a `project.yml` requirement.  Always
+		// keep these.
+		if k.Name != "" {
+			found := false
+
+			for _, r := range keep {
+				if k.Name == r.Name() {
+					found = true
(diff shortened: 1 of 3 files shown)


newt/deprepo/deprepo.go
```diff
@@ -236,15 +236,20 @@ func PruneMatrix(m *Matrix, repos RepoMap, rootReqs RequirementMap) error {
...
-			if r.Name() == k.Name {
-				found = true
-				break
+		// The empty string indicates a `project.yml` requirement.  Always
+		// keep these.
+		if k.Name != "" {
+			found := false
+
+			for _, r := range keep {
+				if k.Name == "" || k.Name == r.Name() {
```
utzig: If `k.Name != ""` in the test above can it ever be `== ""` here?

ccollins476ad: You're right, good catch.

PR #367, opened 2020-01-24
Fix clone of repos for non-[install|upgrade] cmds

utzig:
Add a new function `TryGetOrDownloadProject()` that should be called by commands that want to check out a repo, eg, upgrade/install/sync. Older `TryGetProject()` now just checks that repos exist and bail out with an error if they don't.

### DIFF
--- a/newt/project/project.go
+++ b/newt/project/project.go
@@ -85,10 +85,10 @@ type Project struct {
 	yc ycfg.YCfg
 }
 
-func initProject(dir string) error {
+func initProject(dir string, download bool) error {
 	var err error
 
-	globalProject, err = LoadProject(dir)
+	globalProject, err = LoadProject(dir, download)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


utzig:
@ccollins476ad This might a look a bit ugly so I would be OK with a bit of refactoring! :-) Also I found it weird the amount of exported functions and the whole callchain in `project.go`, maybe there was a reason to have those some time ago but now they only seemed to be used by local code so maybe we could do a bit of "unexport" stuff refactoring here.

newt/project/project.go
```diff
@@ -499,16 +506,24 @@ func (proj *Project) verifyNewtCompat() error {
...
+func (proj *Project) addRepo(r *repo.Repo, download bool) error {
+	if download {
+		if err := r.EnsureExists(); err != nil {
+			return err
+		}
+	} else {
+		if !r.CheckExists() {
+			return util.NewNewtError(
+				fmt.Sprintf(
+					"Repo \"%s\" is not installed, please run `upgrade`!", r.Name()))
```
ccollins476ad: Do you think this should say `newt upgrade`?  I guess the `newt` part is pretty obvious.  Up to you!

utzig: I've found it too long, and I hoped nobody would actually type `$ upgrade` in the shell, but I changed it to show `newt upgrade`, it's probably better!

PR #368, opened 2020-01-27
Add commands to build and browse man pages

utzig:
This adds a PoC level support for `newt` to generate and browse the doxygen in man-page format. No proper error handling, etc, is currently implemented...

When generating man-pages it looks for a file names `man-pages.conf ` in all repos, and if found those are run under doxygen. Generating man-pages:

```sh
$ newt man-build
Preparing man-pages, running doxygen for "apache-mynewt-nimble"
Preparing man-pages, running doxygen for "apache-mynewt-core"
Updating man-page index caches
```

Searching:

```sh
$ newt apropos os_mutex_
os_mutex_test_basic.c (3) - (unknown subject)
os_mutex_test_case_1.c (3) - (unknown subject)
os_mutex_test_case_2.c (3) - (unknown subject)
os_mutex_init (3)    - (unknown subject)
os_mutex_pend (3)    - (unknown subject)
os_mutex_release (3) - (unknown subject)
```

Browse a man page:

```
$ newt man os_mutex_pend
OSMutex(3)                                        Library Functions Manual                                       OSMutex(3)

NAME
       OSMutex

SYNOPSIS
   Data Structures
       struct os_mutex
           OS mutex structure.

   Functions
       os_mutex::SLIST_HEAD (, os_task) mu_head
       os_error_t os_mutex_init (struct os_mutex *mu)
           Create a mutex and initialize it.
       os_error_t os_mutex_release (struct os_mutex *mu)
           Release a mutex.
       os_error_t os_mutex_pend (struct os_mutex *mu, os_time_t timeout)
           Pend (wait) for a mutex.

   Variables
       uint8_t os_mutex::_pad
       uint8_t os_mutex::mu_prio
           Mutex owner's default priority.
       uint16_t os_mutex::mu_level
           Mutex call nesting level.
       struct os_task * os_mutex::mu_owner
           Task that owns the mutex.

Detailed Description
Function Documentation
...
```

The `man` commands loads the page in the given `PAGER` as would be expected...

Requires a `man-pages.conf` for each repo, eg: https://github.com/apache/mynewt-core/pull/2168

PR #370, opened 2020-01-27
Print OneTimeWarning errors to stderr

utzig:
OneTimeWarning currently prints to `stdout`, which when warnings are found, results in test cases run in CI to fail, so update it to print to `stderr`.


utzig:
@ccollins476ad Not sure about possible side-effects of changing this, but the "WARNING: apache-mynewt-core: Repo version missing from compatibility map" message printed to `stdout` is the reason `master`, #367 and #368 are failing on the CI.

PR #371, opened 2020-01-27
Remove deprecated `install` and `sync` commands

ccollins476ad:


utzig:
@ccollins476ad Should the following docs be removed as well: `docs/command_list/newt_install.rst` and `docs/command_list/newt_sync.rst` ?

sjanc:
I'm fine with that but we also need to update all tutorials and web docs we have before this goes into release

PR #373, opened 2020-01-27
Fix some issues reported by Coverity

ccollins476ad:


newt/toolchain/compiler.go
```diff
@@ -1374,7 +1373,7 @@ func (c *Compiler) CopySymbols(infile string, outfile string, sm *symbol.SymbolM
 	if err != nil {
 		return err
 	}
-	return err
+	return nil
```
utzig: I don't understand this issue and the one above it, if `err` is being tested for non-`nil` value above could it be anything different than `nil` in the `return`? (Non related to the question, I guess the test above could be removed for this one). The other changes in this PR look good.

ccollins476ad: @utzig You are right - there was no actual bug here.  Unfortunately I cannot post the actual Coverity scan results since they come from a private version of the analyzer, but the description of the issue is:
> Identical code for different branches

So even though there is no bug, I agree with Coverity that the code did not look right.  I think there are two ways to fix this:
1. Remove the `if err != nil` entirely.  Keep the `return err`.
2. Do what I did (change `return err` to `return nil`).

I prefer option 2 because it is more "Go-like" consistent with the rest of the code.  Also, I think it improves maintainability because it makes it easy to add more code before the `return nil`.

PR #374, opened 2020-01-28
Fix downloading repo dependencies

utzig:
This is a bug left after a previous commit that fixed the download behavirour to only happen for network related commands, like `upgrade`. The repo dependencies was left non-updated at the time, so it would not be possible to resume/retry failures in downloads of dependencies. The bug can also be reproduced by simply removing (`rm -rf`) any repo that is included as dependency of some other repo.

### DIFF
--- a/newt/project/project.go
+++ b/newt/project/project.go
@@ -583,7 +583,7 @@ func (proj *Project) loadConfig(download bool) error {
 	// Read `repository.yml` files belonging to dependee repos from disk.
 	// These repos might not be specified in the `project.yml` file, but they
 	// are still part of the project.
-	if err := proj.loadRepoDeps(false); err != nil {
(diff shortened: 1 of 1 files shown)


utzig:
> (I chose mcuboot at random, but I think it is one of the smaller repos)

```
$ du -sh repos/*                    
430M	repos/apache-mynewt-core
1.4M	repos/apache-mynewt-mcumgr
20M	repos/apache-mynewt-nimble
67M	repos/mcuboot
```

utzig:
> It might be a good idea to add a new step to the travis tests:
> 
> ```
> rm -rf repos/mcuboot
> newt upgrade
> ```

https://github.com/JuulLabs-OSS/mynewt-travis-ci/pull/74

PR #377, opened 2020-02-14
Add support for conditional submodules updates

utzig:
This adds support for repos to inform which submodules are required for a proper build on Mynewt. The main driver for this feature is that the current `ext/mbedtls` on mcuboot is a submodule that is not required by Mynewt builds, and the amount of submodules is going to grow in the future, eg, https://github.com/JuulLabs-OSS/mcuboot/pull/653.

The way this works is `newt` is going to look for a "repo.submodules" key inside the "repository.yml" file. This is a list of submodules to be cloned. One of two things could happen:

1) If the key is no present `newt` is going to clone all submodules (to preserve compatibility).
2) If the key is found, each of the submodules declared will be cloned one by one.

For example, if "repository.yml" has the following contents:

```
repo.submodules: ""
```

In this case submodule cloning is skipped (with a verbose message stating it).

```
repo.submodules:
    - "ext/mbetls"
    - "boot/cypress/somemodule"
```

This will clone the two modules declared but not other submodules that might eventually exist.

This PR also makes cloning submodules recursive, so submodules inside submodules are also cloned.

### DIFF
--- a/newt/repo/repo.go
+++ b/newt/repo/repo.go
@@ -62,6 +62,9 @@ type Repo struct {
 
 	// version => commit
 	vers map[newtutil.RepoVersion]string
+
+	hasSubmodules bool
+	submodules    []string
 }
 
 type RepoDependency struct {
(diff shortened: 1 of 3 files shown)


newt/repo/repo.go
```diff
@@ -191,6 +194,22 @@ func (r *Repo) patchesFilePath() string {
 		"/.patches/"
 }
 
+// Check that there is a list of submodules to clone; if the key is
+// not present, assume that all submodules should be cloned, otherwise
+// only those configured will be cloned (or none if "")
+func (r *Repo) loadSubmoduleConfig(yc *ycfg.YCfg) error {
```
ccollins476ad: Other functions that take a `ycfg.YCfg` object take it by value, not reference.  It might make sense to change this function to match the others.

After programming in Go for several years, I still have no idea when to use pointers and when not to, so I am *not* saying pass-by-value is "better" here, just more consistent :).

utzig: > Other functions that take a ycfg.YCfg object take it by value, not reference. It might make sense to change this function to match the others.

Done, thanks!

PR #382, opened 2020-03-05
Fix skipping the copy of empty dirs on tree copies

utzig:
The CopyDir implementation did not copy empty directories which was resulting in a dirty repo after the tree copy with non-cloned submodules (empty dirs) was performed.

The issue resulted from calling `MkdirAll` with the parent directory of the current directory under copy, followed by a `ReadDir` which would result in an empty list of files/dirs, skipping the directory creation (by skipping the recursive tree copy).

Also removes the mcuboot submodule hack.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -525,7 +525,7 @@ func CopyDir(srcDirStr, dstDirStr string) error {
 		return ChildNewtError(err)
 	}
 
-	if err := os.MkdirAll(filepath.Dir(dstDirStr), info.Mode()); err != nil {
+	if err := os.MkdirAll(dstDirStr, info.Mode()); err != nil {
 		return ChildNewtError(err)
 	}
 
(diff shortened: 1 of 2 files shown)


utzig:
@ccollins476ad This fixes the issue with dirty state in mcuboot directory because of submodules that were not copied. I am not so sure the `MkdirAll` change won't have side-effects, so another easy approach would be to just check that `ReadDir` returns an empty `info` and in this case create the directory. I will wait for your opinion!

util/util.go
```diff
@@ -525,7 +525,7 @@ func CopyDir(srcDirStr, dstDirStr string) error {
 		return ChildNewtError(err)
 	}
 
-	if err := os.MkdirAll(filepath.Dir(dstDirStr), info.Mode()); err != nil {
+	if err := os.MkdirAll(dstDirStr, info.Mode()); err != nil {
```
utzig: Or maybe this should be `filepath.Clean(dstDirStr)` ?

PR #384, opened 2020-03-10
Update go.sum with missing checksums

sjanc:


utzig:
Those updates also happen automatically when I build locally, I am not sure why it doesn't happen for @ccollins476ad, but I am fine with this.

ccollins476ad:
@utzig I am also not sure why the `go.sum` changes are inconsistent.  I have seen this a few other times and never could figure it out.  I think this must be a bug in Go, but haven't verified that.

PR #392, opened 2020-04-01
Update documentation for mfg create

dwld:
Documentation update for the mfg create command in order to match the current implementation.

### DIFF
--- a/docs/command_list/newt_mfg.rst
+++ b/docs/command_list/newt_mfg.rst
@@ -105,7 +105,7 @@ Build the bootloader and app images.
 .. code-block:: console
 
     $ newt build rb_boot
-    $ newt build rb_blink && newt create-image rb_blinky 0.0.1
+    $ newt create-image rb_blinky 0.0.1
 
 Run the ``newt mfg create`` command to collect all the manufacturing snapshot files.
 
--- a/docs/command_list/newt_mfg.rst
+++ b/docs/command_list/newt_mfg.rst
@@ -104,7 +104,7 @@ Build the bootloader and app images.
 
 .. code-block:: console
 
-    $ newt build rb_boot && newt create-image rb_boot 0.0.1
+    $ newt build rb_boot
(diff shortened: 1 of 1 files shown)


docs/command_list/newt_mfg.rst
```diff
@@ -87,16 +87,25 @@ The ``rb_blinky_rsa`` package needs a pkg.yml file. In addition it is needs a mf
...
+    mfg.meta:
+        area: FLASH_AREA_BOOTLOADER
 
 Build the bootloader and app images.
 
 .. code-block:: console
 
-    $ newt build rb_boot
-    $ newt create-image rb_blinky 0.0.1
+    $ newt build rb_boot && newt create-image rb_boot 0.0.1
```
ccollins476ad, line 107: Assuming `rb_boot` is a boot loader target, that should be `newt build rb_boot` (rather than `newt create-image rb_boot 0.0.1`).

`create-image` would also work, but it creates an unnecessary `.img` file with an image header (and it requires you to specify a version number that doesn't get included anywhere).

dwld: @ccollins476ad, thanks for your feedback. I removed it.

docs/command_list/newt_mfg.rst
```diff
@@ -87,16 +87,25 @@ The ``rb_blinky_rsa`` package needs a pkg.yml file. In addition it is needs a mf
...
+    mfg.meta:
+        area: FLASH_AREA_BOOTLOADER
 
 Build the bootloader and app images.
 
 .. code-block:: console
 
     $ newt build rb_boot
-    $ newt create-image rb_blinky 0.0.1
+    $ newt build rb_blink && newt create-image rb_blinky 0.0.1
```
ccollins476ad, line 108: That should be `rb_boot` (not `rb_blink`), right? :)

dwld: Thanks, @ccollins476ad. You are right ;) I am a newbie to newt... Changed.

PR #393, opened 2020-04-07
run: Fix options parsing for newt run

vrahane:
- Fixing parsing of encryption filenames and options for the run command


vrahane:
@ccollins476ad Thanks, done

PR #397, opened 2020-04-30
Add include_dirs setting to syscfg

michal-narajowski:
Allows to specify directories to include in an external SDK
that does not follow the convention used by Newt of putting
everything under "src/ext/".

### DIFF
--- a/newt/builder/buildpackage.go
+++ b/newt/builder/buildpackage.go
@@ -102,7 +102,7 @@ func (bpkg *BuildPackage) recursiveIncludePaths(
 
 	incls := []string{}
 	for _, p := range deps {
-		incls = append(incls, p.publicIncludeDirs(b.targetBuilder.bspPkg)...)
+		incls = append(incls, p.publicIncludeDirs(b)...)
 	}
 
 	return incls, nil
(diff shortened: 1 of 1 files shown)


newt/builder/buildpackage.go
```diff
@@ -251,6 +252,16 @@ func (bpkg *BuildPackage) publicIncludeDirs(bspPkg *pkg.BspPackage) []string {
...
 		sdkIncls := bpkg.findSdkIncludes()
 		incls = append(incls, sdkIncls...)
+
+		settings := b.cfg.AllSettingsForLpkg(bpkg.rpkg.Lpkg)
+
+		inclDirs, err := bpkg.rpkg.Lpkg.PkgY.GetValStringSlice(
+			"pkg.include_dirs", settings)
+		util.OneTimeWarningError(err)
+
+		for _, src := range inclDirs {
```
ccollins476ad, line 262: nitpick: `src` is an odd name to use here.  Something like `dir` feels more appropriate to me.

michal-narajowski: Oops, I missed that. Thanks!

PR #401, opened 2020-05-21
cmake: Make CMakeLists.txt more readable

michal-narajowski:


kasjer:
CMakeLists.txt look much nicer now.
:+1: 
But since you are changing style only maybe you could break lines to reasonable length (optionally).
Having lines with thousands of character is difficult to check.

But this is just a suggestion. It is improvement by itself. Great work as usual.

michal-narajowski:
@kasjer I agree, I want to do that too, but it will require a bit more work as most of those long lines are generated from lists of flags or files. I will add another PR later that changes that.

PR #403, opened 2020-06-07
Execute .sh scripts with bash on windows

jipanienko:
Windows load and debug scripts run .sh scripts with bash starting from windows .cmd script.
around 100 bsp has same .cmd script with the same text @bash "%~dp0%~n0.sh"
if newt started .sh file with bash those .cmd scripts can be deleted.
same command can be used for post build that don't have overwrite as download and debug do.

bash will be hardcoded as /bin/sh is for Linux and NEWT_SH environment can change it to something else if needs to be.

in shorts command lines that start with somthing.sh will be prepend with bash on windows

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -369,8 +369,17 @@ func ShellCommandLimitDbgOutput(
 		name = "/bin/sh"
 		args = []string{"-c", strings.Replace(cmd, "\"", "\\\"", -1)}
 	} else {
-		name = cmdStrs[0]
-		args = cmdStrs[1:]
+		if strings.HasSuffix(cmdStrs[0], ".sh") {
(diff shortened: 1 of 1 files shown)


ccollins476ad:
Thanks @jipanienko, it looks good!  I don't have a Windows machine handy to test this on.  I assume you tried this and it worked for you?

kasjer:
> Thanks @jipanienko, it looks good! I don't have a Windows machine handy to test this on. I assume you tried this and it worked for you?

I use it for pre_build scripts and for download and for debug (.WINDOWS.OVERWRITE removed from bsp)
It's not breaking anything nobody used .sh on Windows yet.

jipanienko:
@kasjer @ccollins476ad tested alrigt

PR #404, opened 2020-06-12
Fix --imgfile option for the `load` command

ccollins476ad:
`--imgfile` lets the user specify the image file to load rather than
letting newt derive it from the target name.

This option wasn't working properly.  For the download scripts to load
the specified file, the `BIN_BASENAME` environment variable needs to be
set to the file's *base* (i.e., strip the `.img` extension).

Prior to this commit, this environment variable wasn't being overridden
at all.  The effect was that the `--imgfile` option was ignored
completely.


ccollins476ad:
Thanks @mlaz.

PR #409, opened 2020-07-05
Add Cflags and Include paths to pre_build_cmds environment

caspermeijn:
Add the CFLAGS and INCLUDE_PATH as used by newt to the environment when
starting a pre_build_cmd. I need this to execute rust bindgen correctly.
This program will parse the mynewt headers and generate rust code from
that and therefore it needs the correct c-flags and include paths.


BrandonArp:
I'm using this branch and it perfectly solves the bindgen problem from Rust. I would love to see this get merged.

PR #410, opened 2020-08-12
newt: Add settings to not execute git on every newt command

michal-narajowski:
Executing git on Windows can be very slow so this PR is
introducing flags in settings that can disable
compatibility verification between Newt and repositories
as well as disable writing repo information in generated headers.


michal-narajowski:
Already fixed by https://github.com/apache/mynewt-newt/commit/69976cfd40db1a095f7837f2f8187aef0d032ca9.

PR #415, opened 2020-09-10
Allow newt to use number of physical cores as default for number of jobs in my newt by default

WeekendSuperhero:
on OS X 10.12 and Big Sur having the jobs default to the number of logical cores slows down build to significantly. This pr changes the default number of jobs to the number of physical cpu cores.


vrahane (approved):
Looks good.

vrahane:
@WeekendSuperhero Can you please clean up the commit messages and squash them into one.

PR #416, opened 2020-10-12
Update newt_linux.rst

MichaelLawton:
Removed references to outdated JuulLabs-OSS package repo, added instructions for downloading and installing binary executable and updated version numbers and console output.

### DIFF
--- a/docs/install/newt_linux.rst
+++ b/docs/install/newt_linux.rst
@@ -1,28 +1,18 @@
 Installing Newt on Linux
 ------------------------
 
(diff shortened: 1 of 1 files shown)


ccollins476ad (approved):
Thanks @MichaelLawton.  The changes look good to me, but do we really want to stop providing the newt debian package?  I guess we haven't been keeping up with the new releases.

I'll approve, but I would like to get someone else's input.  I'll plan to merge in a few days if no one objects.  I suppose we can always change the docs back if we get back into supplying deb packages.

PR #418, opened 2020-10-23
Add Section TLV support to images

agross-korg:
This set of patches adds support for the Section TLV to newt.  The Section TLV is used to include information about an elf section in the binary.  This is necessary in some instances because the distributed images do not have accompanying elf files.  The TLV contains a offset and length to denote the elf section offset and length.

This TLV can be used by other tooling to access the specific section contents of a binary.

### DIFF
--- a/newt/imgprod/imgprod.go
+++ b/newt/imgprod/imgprod.go
@@ -24,6 +24,8 @@ package imgprod
 import (
 	"fmt"
 	"os"
+	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/apache/mynewt-artifact/flash"
@@ -45,6 +47,7 @@ type ImageProdOpts struct {
 	AppHexFilename    string
 	EncKeyFilename    string
 	EncKeyIndex       int
+	Sections          []image.Section
 	Version           image.ImageVersion
 	SigKeys           []sec.PrivSignKey
 	BaseAddr          int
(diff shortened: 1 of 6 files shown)


newt/imgprod/imgprod.go
```diff
@@ -312,10 +318,55 @@ func OptsFromTgtBldr(b *builder.TargetBuilder, ver image.ImageVersion,
 
 func ProduceAll(t *builder.TargetBuilder, ver image.ImageVersion,
 	sigKeys []sec.PrivSignKey, encKeyFilename string, encKeyIndex int,
-	hdrPad int, imagePad int) error {
+	hdrPad int, imagePad int, sectionString string) error {
+
+        elfPath := t.AppBuilder.AppElfPath()
+
+        cmdName := "arm-none-eabi-objdump"
```
ccollins476ad, line 325: This seems like something that should be specified in the compiler package.

I see newt does this same thing in the `size` command, so I can't fault you for doing it here.  Maybe a future improvement to consider later.

ccollins476ad (changes requested):
Looks good.  I just have a few irksome go-related requests:

1. Can you please run `gofmt` on the modified files?  Formatting is not required by the language, but it is pretty much standard anyway.

2. Go uses camelCase rather than snake_case.  This is a convention that is rooted in the rules of the language - identifiers starting with a capital letter are public; the rest are private to the containing package.

aditihilbert:
handy feature to allow additional executable - thanks!

newt/imgprod/imgprod.go
agross-korg: yeah good point.  I'll have to put that on the future todo list.

agross-korg: go fmt comment addressed.  Pushed new set of changes incorporating that change.

nkaje (approved):

agross-korg:
Fixed comments from @ccollins476ad

PR #419, opened 2020-10-30
builder: Add package names to the environment variables

caspermeijn:
This adds a list of the package names to the environment variables. I use this for my rust bindings. I only enable the bindings for packages that are actually enabled.


caspermeijn:
@sjanc I would like this to be merge before the 1.9 release.

sjanc (approved):

PR #420, opened 2020-11-18
newt: Support legacy TLV option

agross-korg:
This patch adds support for generating images using legacy TLV values for
AES_NONCE and SECRET_ID.  This allows newer newt tool versions to be able to
create new application images for devices which only support legacy TLV values.

Bumps mynewt-artifact to v0.0.21

Signed-off-by: Andy Gross <andy.gross@juul.com>


nkaje (approved):

InternetofAwesome:
Is there anything holding this from being merged? 
If not, would someone with privileges mind merging?

utzig (approved):

PR #421, opened 2020-11-23
newt; fix flag use when building dependencies.

mkiiskila:
Specifically, need the missing asflags when using clang to build sim.


mlaz (approved):
LGTM

PR #423, opened 2020-12-04
Relax empty values override restriction

kasjer:
With this change syscfg variables that are defined but don't have
default values in syscfg.defs, can be set in other packages and not only
bsp/app/target.

This way MCU packages can set values that otherwise would have to be
set by every BSP.

Since there is no default value, single package that provides value
is not creating any conflicts.


michal-narajowski (approved):
LGTM

PR #431, opened 2021-03-12
Update go.sum and go.mod for Windows build

sjanc:
Those seem to be required by build on Windows and result in 'dirty'
binary if newt is compiled with by build.sh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,5 @@
 *~
 tags
 newt/newt
+newt/newt.exe
 newtmgr/newtmgr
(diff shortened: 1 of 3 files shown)


utzig:
I wonder if `github.com/StackExchange/wmi` and `github.com/go-ole/go-ole` will just be ignore on other platforms. Have you tried building with those changes on Linux/Mac?

utzig:
Both packages have comments for Windows only, eg: https://github.com/StackExchange/wmi/blob/master/wmi.go#L1. I think it should not be a problem.

sjanc:
yes, I've check on linux and on mac

PR #440, opened 2021-07-12
mfg: account for updated bootloader path in mfg command

vikrant-proxy:
Bootloader binaries seem to have moved to `app/@mcuboot`, which causes `mfg` command to fail, see https://github.com/apache/mynewt-newt/issues/439. This PR attempts to account for this change.


vrahane (approved):
This does look fine to me. I will let it get reviewed by others as well since we do not want this to be a compatibility breaking change for others.

PR #441, opened 2021-07-12
mfg: Add a base address field 

vikrant-proxy:
Chips with > 1 core, like `nrf53` have the second core mapped to a different address offset than 0x0 (in the case of nrf53, network core is based at 0x1000000). This PR adds a base address field that can be provided in the `mfg.yml` in an attempt to get `mfg` to create appropriately sized binaries. Without this change, the manufacturing image tends to become very large in size (several megabytes) as the base address for that core is not 0x0.
This PR introduces an optional `base-address` command line option that can be provided as an input to `mfg create`. This value will default to 0 internally if this field is not provided.

### DIFF
--- a/newt/cli/mfg_cmds.go
+++ b/newt/cli/mfg_cmds.go
@@ -31,6 +31,8 @@ import (
 	"mynewt.apache.org/newt/util"
 )
 
+var baseAddress int
+
 func ResolveMfgPkg(pkgName string) (*pkg.LocalPackage, error) {
 	proj := TryGetProject()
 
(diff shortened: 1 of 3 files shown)


vrahane:
This also contains changes from https://github.com/apache/mynewt-newt/pull/440, I think those should be removed from here and mention in the description that this needs to be merged after https://github.com/apache/mynewt-newt/pull/440.

vrahane:
I think this change is reasonable, it is required for nrf53 netcore for sure. Wonder if there is a way to make this mandatory somehow for BSPs with dual core instead of making it optional.

vrahane:
cc @ccollins476ad @utzig @aditihilbert @agross-korg @nkaje, making sure mfg images are not affected for dialog with this change.

vrahane:
We should probably also add command line option for it: `--baseaddress=0x0`

vrahane (approved):
LGTM, let's wait for other's approval.

PR #444, opened 2021-08-30
Fix copying repos with symlinks

sjanc:
The CopyDir implementation did not created symlinks but instead was
creating copy of a files resultign in dirty repo.


utzig (approved):
LGTM

kasjer (approved):
Verified on Windows Git Bash

PR #453, opened 2022-02-17
Skip include dirs that do not exist

andrzej-kaczmarek:
It does not really matter for build process, but makes debugging build
issues easier as it does not clutter command line with tons of unused
include directories.


kasjer (commented):
+1

sjanc (approved):

PR #454, opened 2022-02-17
 Allow to inject syscfg from command line

andrzej-kaczmarek:
This adds new -S/--syscfg command line switch to build and config
commands which allows to inject syscfg settings from command line.
Injected settings have highest priority and thus will override any
other value set by any other package.

The purpose of this setting is to easily allow to use single targets
to build multiple similar images, e.g. that have only different BLE
public address different. It is not intended to be used instead of
having multiple targets to build different images (which is still
preferred setup), but to allow quickly build variants of the same
target wihout need to edit files every time.


andrzej-kaczmarek:
this is pushed on top of https://github.com/apache/mynewt-newt/pull/452

PR #457, opened 2022-03-07
Add dependencies for sysinit functions

andrzej-kaczmarek:
Currently it's only possible to specify an arbitrary stage number for
sysinit function. This can be a bit troublesome if we need to assure
some packages are initialized in specific order since it required to
adjust stages for all relevant packages. Also some packages have stages
defined via syscfg (which is horrible) which means they can be changed
and the whole setup breaks.

This patch allows to order sysinit by dependencies between functions.
New syntax is allowed in addition to stage number:
- $after:func_name
- $before:func_name

The former will ensure that function is called after "func_name", and
the latter - after "func_name".

Each sysinit function shall specify one stage number or any number of
dependencies - those cannot be mixed. Some examples of valid sysinit
definitions:

(1)
func_1: 100

(2)
func_1: $before:func_2
func_2: $after:func_1

(3)
func_1:
  - $before:func_2
  - $after:func_3
func_2: $before:func_4,$before:func_5

Note that (2) is ok even though both functions depend on each other,
since effectively both dependencies are the same.

Multiple dependencies can be specified as a list or as a comma-separated
string (3).

An error will be raised if circular dependency is found when resolving
order of functions.


andrzej-kaczmarek:
sysinit

sjanc (approved):

PR #466, opened 2022-03-14
Do not enclose syscfg string values in parentheses

andrzej-kaczmarek:


kasjer (approved):
Looks and works as advertised, thanks.

PR #472, opened 2022-03-30
Add options to skip some git related stuff on each build

andrzej-kaczmarek:
This adds two new options to newtrc.yml:
- skip_newt_compat - allows to skip checking newt compatibility
- skip_syscfg_repo_hash - allows to skip generation of repo hashes to
  syscfg

Those can skip operations that are not essential for build but can be
potentially git-intensive which means they are painfully slow on
Windows.

Both options are boolean and disabled by default.


kasjer (approved):
Works on Windows without executing git

PR #479, opened 2022-06-08
flashmap: Define more symbols in sysflash.h

andrzej-kaczmarek:
sysflash.h will now have more symbols defined for each flash area.


kasjer (approved):
+2, tested OK

PR #483, opened 2022-06-29
mfg: Fix mfg app name

andrzej-kaczmarek:
This reverts commit 67de80e.


vrahane (approved):
Looks good. @vikrant-proxy is not using mynewt right now so, I don’t think he will be able to test the changes. We are good as long as it works fine with nrf5340 builds.

sjanc (approved):

PR #490, opened 2022-09-27
builder: Allow to override compiler in target

andrzej-kaczmarek:
This allows to override compiler packge in target using "target.compiler" key.


kasjer (approved):
LGTM

sjanc (approved):

PR #497, opened 2023-01-02
Allow to download external repos without project.yml file

m-gorecki:
This allows to download external repos like SDK packages, which do not contain project.yml files. Now we do not have to keep all of the SDKs in apache-mynewt-core repository. This also allows to use "@<repo-name>" syntax in the pkg.include_dirs and pkg.src_dirs fields to add source/include files from specified folders from external repos.

### DIFF
--- a/newt/repo/repo.go
+++ b/newt/repo/repo.go
@@ -194,6 +194,18 @@ func (r *Repo) patchesFilePath() string {
 		"/.patches/"
 }
 
+// Checks for repository.yml file presence in specified repo folder.
+// If there is no such file, the repo in specified folder is external.
+func (r *Repo) IsExternal(dir string) bool {
+	path := dir + "/repository.yml"
+	_, err := os.Stat(path)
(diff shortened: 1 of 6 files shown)


kasjer (approved):
LGTM

newt/repo/repo.go
```diff
@@ -194,6 +194,18 @@ func (r *Repo) patchesFilePath() string {
...
+// Checks for repository.yml file presence in specified repo folder.
+// If there is no such file, the repo in specified folder is external.
+func (r *Repo) IsExternal(dir string) bool {
+	path := dir + "/repository.yml"
+	_, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return true
+	} else {
+		return false
+	}
```
michal-narajowski, line 206:
```suggestion
	return os.IsNotExist(err)
```

michal-narajowski (approved):
LGTM

PR #504, opened 2023-01-30
builder: Allow to specify source files for package

andrzej-kaczmarek:
This adds consistent naming for pkg settings to include/ignore source
files and directories in pkg.yml:
- source_dirs
- ignore_dirs
- ignore_files

Old names still work for backwards compatibility, but only if new
variant is not used.


kasjer (approved):
Works for me

PR #506, opened 2023-02-16
compiler: Split long ar command lines into separate commands

andrzej-kaczmarek:
This is an ugly hack to deal with command line length limitation on Windows. It will split long ar invocation into series of shorter ones so Windows can execute them properly.

Windows supports command line length up to 32k but we'll use 30k just so we don't have to worry about some corner cases.


kasjer (approved):
Fixes mynewt builds with LVGL on Windows where number of objects is substantial.

PR #507, opened 2023-02-20
Allow to add external repos from pkg.yml

m-gorecki:
This allows to add exernal repositories to pkg.yml files. Repositories will be downloaded when "newt upgrade" command is called. The syntax of adding new repo to the pkg.yml file is the same as adding new repo to the project.yml file.


kasjer (approved):
Works as advertised.

andrzej-kaczmarek (approved):

PR #510, opened 2023-02-27
compiler: Add compiler flags $(...) patterns 

m-gorecki:
This adds the possibility to define and use $(...) patterns with
compiler flags. To do so you have to define new value in compiler's syscfg.yml
file, for example:

EXAMPLE_FLAGS:
  value:"-my -compiler -flags"

Expression "$(EXAMPLE_FLAGS)" used in any compiler flags definition in compiler.yml will
be now replaced with the defined value (in this case whith "-my -compiler -flags").
It is also possible to override or add new patterns from other syscfg.yml files, for example
from some application's syscfg.yml

### DIFF
--- a/newt/toolchain/compiler.go
+++ b/newt/toolchain/compiler.go
@@ -90,6 +90,7 @@ type Compiler struct {
 	baseDir               string
 	srcDir                string
 	dstDir                string
+	settings              *cfgv.Settings
 
 	// The info to be applied during compilation.
 	info CompilerInfo
@@ -250,7 +251,7 @@ func (ci *CompilerInfo) AddCompilerInfo(newCi *CompilerInfo) {
 }
 
 func NewCompiler(compilerDir string, dstDir string,
-	buildProfile string) (*Compiler, error) {
+	buildProfile string, cfg *cfgv.Settings) (*Compiler, error) {
 
 	c := &Compiler{
 		mutex:           &sync.Mutex{},
@@ -267,7 +268,7 @@ func NewCompiler(compilerDir string, dstDir string,
 	util.StatusMessage(util.VERBOSITY_VERBOSE,
 		"Loading compiler %s, buildProfile %s\n", compilerDir,
 		buildProfile)
-	err := c.load(compilerDir, buildProfile)
+	err := c.load(compilerDir, buildProfile, cfg)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


newt/toolchain/compiler.go
```diff
@@ -152,6 +153,36 @@ func (c *Compiler) GetLocalCompilerInfo() CompilerInfo {
 	return c.lclInfo
 }
 
+func (c *Compiler) ReplaceWildcardFlags() {
+	for _, wildcard := range c.settings.Names() {
+
+		for i, flag := range c.lclInfo.Lflags {
+			if strings.Contains(flag, "$("+wildcard+")") {
+				c.lclInfo.Lflags[i] = strings.ReplaceAll(flag, "$("+wildcard+")", c.settings.Get(wildcard))
```
andrzej-kaczmarek, line 161: would be better to match all `$(...)` in string and then replace each with corresponding syscfg value. this also allows to emit an error if referenced syscfg does not exist.

newt/toolchain/compiler.go
```diff
@@ -260,6 +291,7 @@ func NewCompiler(compilerDir string, dstDir string,
 		dstDir:          dstDir,
 		extraDeps:       []string{},
 		compileCommands: []CompileCommand{},
+		settings:        cfg,
```
andrzej-kaczmarek, line 294: pass directly to `load()`, no need to store it if not used in other places

newt/toolchain/compiler.go
```diff
@@ -332,6 +364,10 @@ func (c *Compiler) load(compilerDir string, buildProfile string) error {
 	c.lclInfo.Lflags = loadFlags(yc, settings, "compiler.ld.flags")
 	c.lclInfo.Aflags = loadFlags(yc, settings, "compiler.as.flags")
 
+	if c.settings != nil {
+		c.ReplaceWildcardFlags()
```
andrzej-kaczmarek, line 368: `loadFlags()` can handle it

andrzej-kaczmarek (changes requested):
commit description is confusing - there are no wildcards used here, it just replaces syscfg referenced using `$(syscfgname)` with their value in compiler settings.

newt/toolchain/compiler.go
```diff
@@ -292,10 +293,20 @@ func loadFlags(yc ycfg.YCfg, settings *cfgv.Settings, key string) []string {
 		}
 	}
 
+	if cfg != nil {
+		for i, flag := range flags {
+			for _, c := range cfg.Names() {
+				if strings.Contains(flag, "$("+c+")") && cfg.Exists(c) {
+					flags[i] = strings.ReplaceAll(flag, "$("+c+")", cfg.Get(c))
```
andrzej-kaczmarek, line 300: no need to iterate all syscfg names to find matches, use `ReplaceAllStringFunc` instead

andrzej-kaczmarek (approved):
lgtm

sjanc (approved):

PR #512, opened 2023-03-03
builder: Allow to override some bsp settings in target

andrzej-kaczmarek:
Target can now override some settings defined in BSP:
- bsp.compiler
- bsp.image_offset
- bsp.image_pad
- bsp.flash_map
- bsp.linkerscript
- bsp.part2linkerscript

To override, simply define the same key in target.yml and it will take precedence over the one defined in bsp.yml.


kasjer (approved):
LGTM. Tested custom flash map and linker scripts

PR #520, opened 2023-05-30
Add a way to set linker flags --whole-archive and --no-whole-archive per pkg

jocelynmass:
Depending on the project, it might be useful to have the ability to link a package surrounded by "--whole-archive" and "--no-whole-archive" flags. 
I tweaked the code to allow pkg.whole_archive option to be set to true or false. I'm completely new to Goland so the code I made is most likely garbage. I just submit this PR with the hope that someone with more skills would take over this and fully make it to the next newt release.


sjanc:
sorry for late reply, looks OK to me!

@m-gorecki could you look at this?

m-gorecki:
LGTM

kasjer:
@jocelynmass hi Jocelyn, your change seems ok and useful there are some conflicts due to late review. Would you care to rebase is it removing conflicts? If you don't have time I can do it. In this case I will force push resolved conflicts and it will be merged soon.
Thanks for contribution and I apologize for late review.

sjanc (approved):

PR #521, opened 2023-06-19
syscfg: Write defs for each APIs provided by included packages

andrzej-kaczmarek:
This adds symbols in syscfg.h for each API provided by packages included in build. The symbol name has "MYNEWT_API_" prefix followed with sanitized API name, all illegal characters in resulting name are replaced with "_".

This allows to check if any package included in build provides specific API. Could be useful if some API is not strictly required, but we can use it if provided by some package.

### DIFF
--- a/newt/syscfg/syscfg.go
+++ b/newt/syscfg/syscfg.go
@@ -1599,7 +1599,19 @@ func writePackages(lpkgs []*pkg.LocalPackage, w io.Writer) {
 	}
 }
 
-func write(cfg Cfg, lpkgs []*pkg.LocalPackage, w io.Writer) {
+func writeApis(apis []string, w io.Writer) {
+	for i, api := range apis {
+		apis[i] = cfgPkgIllegalChar.ReplaceAllLiteralString(api, "_")
+	}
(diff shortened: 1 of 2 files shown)


newt/syscfg/syscfg.go
```diff
@@ -1599,7 +1599,19 @@ func writePackages(lpkgs []*pkg.LocalPackage, w io.Writer) {
...
-func write(cfg Cfg, lpkgs []*pkg.LocalPackage, w io.Writer) {
+func writeApis(apis []string, w io.Writer) {
+	for i, api := range apis {
+		apis[i] = cfgPkgIllegalChar.ReplaceAllLiteralString(api, "_")
+	}
+	sort.Strings(apis)
+
+	fmt.Fprintf(w, "/*** Included APIs */\n")
+	for _, name := range apis {
+		fmt.Fprintf(w, "#define MYNEWT_API_%s 1\n", name)
```
andrzej-kaczmarek, line 1610: `MYNEWT_API` -> `MYNEWT_PKG_API`?

PR #522, opened 2023-06-20
Fix CmakeList.txt $(...) compiler flags

m-gorecki:
This fixes the issue: https://github.com/apache/mynewt-newt/issues/516


kasjer (approved):
LGTM, tested without problem with both linker and compiler flags

PR #523, opened 2023-07-26
compiler: Fix pkg.source_files compiler flags

m-gorecki:
This fixes the issue: https://github.com/apache/mynewt-newt/issues/517


kasjer (approved):
LGTM. Verified to work

PR #527, opened 2023-08-21
Enable upgrading only specified repos

utzig:
Reimplement support for `newt upgrade <repo1> <repo2>...`. Not specifying repos upgrades everything; specifying repos update only the ones given, without upgrading/installing its dependencies, etc.

This should allow advanced users to install only what is required, eg. if developing an app for nrf52x one could simply run:

`newt upgrade --depth=1 apache-mynewt-core apache-mynewt-nimble nordic-nrfx`

This is also useful for CI, eg, for MCUboot it's possible to use MCUboot as the root project and run:

`newt upgrade --depth=1 apache-mynewt-core nordic-nrfx mbedtls`

And this brings down the complete download from 1GB to just around 250MB.

### DIFF
--- a/newt/cli/project_cmds.go
+++ b/newt/cli/project_cmds.go
@@ -121,14 +121,14 @@ func makeRepoPredicate(repoNames []string) func(r *repo.Repo) bool {
 }
 
 func upgradeRunCmd(cmd *cobra.Command, args []string) {
-	proj := TryGetOrDownloadProject()
+	proj := TryGetOrDownloadProject(args)
 	interfaces.SetProject(proj)
 
-	proj.GetPkgRepos()
+	proj.GetPkgRepos(args)
 
 	pred := makeRepoPredicate(args)
 	if err := proj.UpgradeIf(
-		newtutil.NewtForce, newtutil.NewtAsk, pred); err != nil {
+		newtutil.NewtForce, newtutil.NewtAsk, args, pred); err != nil {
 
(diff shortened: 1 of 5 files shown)


utzig:
Hmmm OK, CI passed now because I fixed the dependency graph when passing `repo` arguments to `upgrade`. So I'll drop the CI changes and it will break again, because there are `repo` arguments in the calls, even though usage is broken and `upgrade` totally ignores them. This probably worked back in the day when there was `newt install`. So, I'll have to end up adding a `--no-deps` parameter or something similar and which shall maintain the broken compatibility with the broken usage...

andrzej-kaczmarek:
we also had idea to add include/exclude repo options to project.yml. this way project can be configured to only include certain repositories required for that project or exclude those not needed.

nevertheless, this one is also nice to have.

utzig:
> we also had idea to add include/exclude repo options to project.yml. this way project can be configured to only include certain repositories required for that project or exclude those not needed.
> 
> nevertheless, this one is also nice to have.

Cool, I've found some regressions when last testing this, which is the reason for the WIP. I'll try to get back to this today or tomorrow, and will leave a comment once I find is is good for review.

vrahane:
@andrzej-kaczmarek and @sjanc  does this PR do the same with dependent/ignore repos as the most recent commits ?

sjanc:
yeap, I think it serves same purpose

PR #531, opened 2023-09-19
Update mod.go to use mynewt-artifact v0.0.25-0.20230515081815-85acad353839

benmccrea:
Use latest mynewt-artifact which includes a fix for images that use HW encryption (commit 54b7160).
Also ran `go mod tidy` to clean up dependencies.


vrahane (approved):
Tested this. Works fine. We can probably do an official release later but for now this change seems good. Waiting for others to approve it.

sjanc (approved):

PR #534, opened 2023-11-07
Add possibility to define global cflags in packages

m-gorecki:
This adds possibility to use global cflags in packages. Until now only cflags defined in app package were used globally while compiling. Now it is possible to add such flags in every package by using "app.cflags" field in package's yaml file. This also adds new command to show which packages within the target contain those global flags.

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -635,11 +635,39 @@ func buildWorker(
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sjanc (approved):

kasjer (changes requested):
while app.cflags is passed to build it is not put in generated CMakelist.txt

kasjer (approved):
Tested OK

PR #538, opened 2024-01-04
Fix sysinit test

sjanc:
Remove unneeded commands and update test reference data.


sjanc:
for now just update test data, but we should reconsider on how we test this and how stable this should be

andrzej-kaczmarek (approved):

PR #540, opened 2024-01-19
newt: Add possibility to use generated linker script

m-gorecki:
Now if "bsp.linkerscript" has value "autogenerated", default linker script
will be passed to the linker.


kasjer (approved):
LGTM, verified to work

newt/pkg/bsp_package.go
```diff
@@ -103,14 +113,33 @@ func (bsp *BspPackage) resolveLinkerScriptSetting(
...
+			if val == "autogenerated" {
+				path, err := bsp.getAutogeneratedLinkerScriptPath()
+				if err != nil {
+					return nil, util.PreNewtError(err,
+						"Could not resolve autogenerated linker script path for target \"%s\"",
+						bsp.yov.Pkg.Name())
+				}
+				paths = append(paths, path)
+				continue
+			}
```
andrzej-kaczmarek, line 141: autogenerated scripts should be only handled if single script is set to "autogenerated", here it should better just return an error

andrzej-kaczmarek (changes requested):

andrzej-kaczmarek (approved):

PR #541, opened 2024-02-13
newt: Fix autogenerated linker script usage

m-gorecki:
PreNewtError instead of NewNewtError was causing crashes. Also
"autogenerated" key should be handled when there is only one
value ("autogenerated") inside the vals slice.


kasjer (approved):
Tested OK

sjanc (approved):

PR #542, opened 2024-02-19
wip: generate linker sections

m-gorecki:
Adding pkg.sections field with sections names will result in generation of two header files ld_ram.h and ld_text.h. Files are used in generated linker scripts.


sjanc (approved):

andrzej-kaczmarek:
how does this work if build contains more than 1 package using this feature?

m-gorecki:
@andrzej-kaczmarek Entries from each pkg used in the build will be collected and placed in the generated header.

andrzej-kaczmarek:
so I don't really understand what's the purpose of that if everything is put in single and thus can only be included in single section of linker script...

kasjer:
@andrzej-kaczmarek It's put in one section (for cortex it's in .text just before .rodata) but it adds `__<section>__start__` and `__<section>__end__` that can be referred in code making link time tables

PR #543, opened 2024-02-23
newt: Fix shallow upgrade with specific commit version

m-gorecki:
Upgrading project using shallow option was causing an error if repository version was specified as specific commit. Because this specific commit could not yet be fetched it could not be found during version validation.
Now if shallow upgrade is being performed, instead of returning an error immediately after commit was not found, we first try to fetch the specified commit.

### DIFF
--- a/newt/repo/version.go
+++ b/newt/repo/version.go
@@ -126,6 +126,15 @@ func (r *Repo) VersionIsValid(ver newtutil.RepoVersion) bool {
 	}
 
 	cs, _ := r.downloader.CommitsFor(r.Path(), ver.Commit)
+
+	// Try to fetch commit if it was not found
+	if len(cs) <= 0 {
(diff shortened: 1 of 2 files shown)


newt/repo/version.go
```diff
@@ -126,6 +126,15 @@ func (r *Repo) VersionIsValid(ver newtutil.RepoVersion) bool {
 	}
 
 	cs, _ := r.downloader.CommitsFor(r.Path(), ver.Commit)
+
+	// For shallow copy commit might not be fetched yet
```
andrzej-kaczmarek, line 130: this could be actually done regardless of whether repo is shallow or now
the same problem exists if user specifies a commit that is not preset on any remote branch, e.g. a sha for PR head

sjanc (approved):

PR #544, opened 2024-02-23
newt: Upgrade repos specified in project and package only once

m-gorecki:
Now repos that are specified in both project.yml and pkg.yml will be upgraded only once - to the version specified in project.yml


kasjer (approved):
Tested OK

PR #545, opened 2024-03-06
newt/compiler: Pass autogenerated linker script's include dir to linker

m-gorecki:
This tells the linker where to look for includes related with autogenerated linker scripts.


kasjer (approved):
Tested OK

PR #546, opened 2024-03-08
newt: Don't stop upgrade when some repos are in dirty state

m-gorecki:
Now if force flag is not set, only repos in dirty state will not get upgraded. The rest of the repos will be upgraded.


kasjer (approved):
LGTM, tested OK

andrzej-kaczmarek (approved):

PR #548, opened 2024-03-11
newt: Fix crash when package is missing

m-gorecki:
When package from build was missing the b.Build() was returning an error before AppBuilder was created. It was resulting in dereferencing null pointer in such situation.


kasjer (approved):
Tested OK

PR #550, opened 2024-03-27
newt: Add external repos patches support

m-gorecki:
Now if package that adds external repository contains directory patches/*external_repo_name*, newt will try to apply patches inside this directory to the external repo.

### DIFF
--- a/newt/cli/project_cmds.go
+++ b/newt/cli/project_cmds.go
@@ -131,6 +131,7 @@ func upgradeRunCmd(cmd *cobra.Command, args []string) {
 	interfaces.SetProject(proj)
 
 	proj.GetPkgRepos()
+	proj.SetGitEnvVariables()
 
 	pred := makeRepoPredicate(args)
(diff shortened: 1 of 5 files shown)


kasjer (approved):
Tested OK.

Check fails because code tries to commit (apply patches) without settings git user.name user.email

sjanc (changes requested):
I think we should use GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL to avoid requiring git configuration by user. Also this would allow us to have 'newt' as committer ie:
GIT_COMMITTER_NAME=newt
GIT_COMMITTER_EMAIL=dev@mynewt.apache.org

sjanc (approved):

kasjer:
When patch fails affected git repository stays in state that `git am --abort` or `git am --skip` are required. Maybe newt could call `git am --abort` as not to leave git in the middle of git am session

kasjer:
Verified to work

PR #554, opened 2024-04-16
builder: Add private includes only to specific packages

m-gorecki:
Instead of adding private includes paths to every package we should add them to the specific packages


kasjer (approved):
LGTM, fixes build of blemesh_models_example_2 when bootutils are included by bsp or startup code.
common.h present in application src folder does not conflict with mbedtls common.h include

sjanc (approved):

PR #556, opened 2024-05-13
Allow overriding MYNEWT_DOWNLOADER from command line

m-gorecki:
This adds new flag to load and run commands.
The flag (--loader or short -L) allows to override the MYNEWT_DOWNLOADER value from syscfg. Usage example:

--loader stm32_programmer_cli

### DIFF
--- a/newt/cli/build_cmds.go
+++ b/newt/cli/build_cmds.go
@@ -347,6 +347,10 @@ func loadRunCmd(cmd *cobra.Command, args []string) {
 		NewtUsage(cmd, util.NewNewtError("Must specify target"))
 	}
 
+	if util.InjectSyscfg != "" {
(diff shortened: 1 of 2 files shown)


andrzej-kaczmarek:
`-S` can be already used to override syscfg from command line

m-gorecki:
In case of loading, there is no use overriding other configs than MYNEWT_DOWNLOADER. I assumed it would be better to simply allow user to override only this specific syscfg value. Under the hood it uses the same code to override the value as `-S` flag, it's just wrapped to not cause confusion

andrzej-kaczmarek:
my main concern is that this simply overrides syscfg defined in some package - that doesn't sound like a maintainable solution.

would be better if we "simply" add loader packages as a generic solution, i.e. add `pkg.type: loader` and move all scripts there. bsp would then specify default and/or supported loader(s) which could be overridden from command line.

PR #557, opened 2024-05-13
Add ignored repositories list

m-gorecki:
This allows to specify a list of repositories names that will be ignored by upgrade command. The list should be placed inside project.yml file and look for example like this:

project.repositories.ignored:
    - apache-mynewt-nimble
    - stm-cmsis_device_f3
    - stm-stm32f3xx_hal_driver

apache-mynewt-core repository can't be ignored.

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -940,18 +940,9 @@ func (b *Builder) CleanArtifacts() {
 	}
 }
 
-func Contains(elements []string, val string) bool {
-	for _, s := range elements {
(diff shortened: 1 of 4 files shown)


kasjer (approved):
Tested OK

andrzej-kaczmarek:
would be nice to also have list of allowed repositories, i.e. I don't want any new repository unless explicitly allowed

andrzej-kaczmarek (approved):

PR #559, opened 2024-05-17
Fix whole archive compiler flag handling

vrahane:


sjanc (changes requested):
please explain what exactly is fixed here in commit message

vrahane:
This does not fix anything. We were seeing some image hash changes on subsequent builds but that was because of something else. Initial thought was this caused it but it didn’t.

PR #566, opened 2024-08-19
Report ambiguous syscfg configuration

m-gorecki:
This addresses: https://github.com/apache/mynewt-newt/issues/565

### DIFF
--- a/.github/targets/ambiguous_fail/pkg.yml
+++ b/.github/targets/ambiguous_fail/pkg.yml
@@ -1,0 +1,5 @@
+pkg.name: "targets/ambiguous_fail"
+pkg.type: target
+pkg.description: 
+pkg.author: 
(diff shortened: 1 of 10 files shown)


sjanc (approved):

sjanc:
would be good to also add unit test that will validate this before closing issue

sjanc (commented):
this also needs to validate if value differs as it is fine to set it to same value from multiple locations

sjanc:
seems to be crashing on specific target
https://github.com/apache/mynewt-nimble/actions/runs/10739293391/job/29840180220?pr=1851

(can be reproduced with .github/nordic_pca10028_blehci target from nimble repo)

m-gorecki:
This should fix the issue: https://github.com/apache/mynewt-newt/pull/570

PR #568, opened 2024-09-06
newt: Fix mfg crash with generated linker scripts

m-gorecki:
This prevents panic error while using newt mfg create with bsps that use autogenerated linker scripts. No yov in bsp means that mfg create command was used (all other commands create bsp structure that contains yov), which means that getting autogenerated linker script path is not needed.


jfclere:
the PR fixes the crash and hex boot and run the app as expected.

sjanc (approved):

PR #575, opened 2024-10-25
Make repos defined in project.yml implicitly allowed

andrzej-kaczmarek:
Repositories defined in project.yml should be allowed by default (even if omitted on allowed list) unless ignored explicitly.


andrzej-kaczmarek:
With this fix, if you only want to allow repositories in `project.yml` you only need to include e.g. `apache-mynewt-core` on the list and all other repos will be allowed implicitly. The single entry is required as otherwise empty allowed list will allow all repos.

Alternatively if you want to allow some extra repos not included in `project.yml` (e.g. defined by packages) its enough to only include those repos on the list as repos defined in `project.yml` will be added implicitly anyway.

m-gorecki (approved):

sjanc (approved):

PR #577, opened 2024-10-28
Add support for SmPL overlay patches

MariuszSkamra:
This adds support for SmPL cocci patches.


sjanc (commented):
I think this should be optional dependency,  if coccinelle newt should just print warning and skip patch

andrzej-kaczmarek:
Perhaps a more portable option would be to have .cocci in separate directory so they can be used to generate .patch manually in case library is updated. In such case .patch will be applied by newt.

sjanc:
another problem is that coccinnelle is not available on windows

PR #586, opened 2024-11-21
Fix CMake project name

andrzej-kaczmarek:
This fixes project name to use both project name and target name to avoid confusion in case there are multiple projects with the same target name (e.g. CLion would list them with the same name on recent projects list).


kasjer (approved):
LGTM, works great

PR #588, opened 2024-11-25
builder: Print load commands output on default

m-gorecki:
Now load command output will be printed on default. To disable this add "hide_load_output: 1" to ~/.newt/newtrc.yml


kasjer (approved):
Works for me

sjanc (approved):

PR #589, opened 2024-11-26
builder: Add conditional flash map overrides in bsp.yml

m-gorecki:
This allows to conditionally override flash maps in bsp.yml and target.yml with bsp.flash_map.<CONDITION>
Initial priority of bsp.yml and target.yml is preserved - if there is any bsp.flash_map definition in target.yml, all flash maps from bsp.yml are ignored. Default bsp.flash_map definition (without any condition) is still needed.


kasjer (approved):
Verified to work, LGTM

PR #591, opened 2024-12-20
newt: Switch resolution warnings from debug to warn log lvl

m-gorecki:
Now newt warns about undefined settings on default


kasjer (approved):
LGTM

sjanc (approved):

PR #593, opened 2025-02-20
newt: Add --list flag to test command

m-gorecki:
Now it's possible to show all available unit tests


sjanc (approved):

sjanc (commented):
Would be nice to update shell completion to also include full test name since it seems to complete only test group

eg 
@apache-mynewt-core/sys/log/full

while --list shows
@apache-mynewt-core/sys/log/full/selftest/fcb2_bookmarks
@apache-mynewt-core/sys/log/full/selftest/perlogidx
and more

BTW   list could be sorted so that test from same package are listed together

PR #595, opened 2025-04-11
newt: If no dependencies are present, still try to build

vrahane:
- If no dependencies are present, still try to build.
- We saw that with no dependencies, just using newt for building without the OS was failing.
- This helped but I am still trying to understand if its the correct solution,  please try it out.


vrahane:
Seems like the solution works but in only one case, `newt upgrade` probably breaks. Need to find a better approach to this.

newt/sysinit/sysinit.go
```diff
@@ -181,6 +180,10 @@ func ResolveStageFuncsOrder(sfs []stage.StageFunc) ([]stage.StageFunc, error) {
 	// direct dependencies between each node of stage X to each node of
 	// stage Y to make sure they can be resolved properly and reordered
 	// if needed due to other dependencies.
+	if stages == nil {
+	  util.FmtNewtError("Cannot resolve sysinit order: no stages defined.")
+		return []stage.StageFunc{}, nil
+	}
```
andrzej-kaczmarek, line 186: I think this can only happen if `sfs` is empty so you can basically check that at the very beginning of the function. Also no need to emit warning.

newt/repo/repo.go
```diff
@@ -198,7 +198,11 @@ func (r *Repo) Name() string {
 }
 
 func (r *Repo) Path() string {
-	return r.localPath
+	if r != nil {
+		return r.localPath
+	} else {
+		return ""
+	}
```
andrzej-kaczmarek, line 205: This will "hide" all issues where `Path()` is called unintentionally on `nil` object, so better to actually explicitly handle this on caller side. Of course as long as there is some valid case in which we can actually have `nil` repo that we should handle properly, not sure when this could happen though tbh.

PR #598, opened 2025-06-02
newt/build: Workaround for whole-archive flag

m-gorecki:
Flag whole-archive is not supported on Mac and it was causing CI failures. Now instead of using this flag with .a file we just pass a text file with list of all object files from package to the linker. So now it looks like this:
@pkg.list
instead of:
--whole-archive pkg.a --no-whole-archive

Passing list of object files will result in pulling unreferenced object code just like in case of whole-archive flag and should work with all compilers.

### DIFF
--- a/newt/builder/build.go
+++ b/newt/builder/build.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


kasjer:
This seems to work for arm builds and pic32 (except for Windows). Mac not tested by me.

kasjer (approved):
Code not inspected, just tested to verify that it works

PR #601, opened 2025-06-27
builder: Fix false positive flash map errors

m-gorecki:
To allow conditional overrides of flash map it's necessary to resolve target, because otherwise the syscfgs that determine the choice of a different flash map will not be available. The problem was that every resolving was trying to detect errors in configuration (also in flash map). As the resolving had to be done before bsp package reloading, while detecting errors the default flash map is loaded. If this default flash map did not define some flash areas used in the target configuration, an error was reported even if the flash map that should override the default one was correct.

Now we don't try to detect error while doing this particular resolving. Instead we do this after reloading bsp package so the properly chosen flash map is used in error detection.

### DIFF
--- a/newt/builder/targetbuild.go
+++ b/newt/builder/targetbuild.go
@@ -95,13 +95,15 @@ func NewTargetTester(target *target.Target,
 		injectedSettings: cfgv.NewSettings(nil),
 	}
 
-	if err := t.ensureResolved(); err != nil {
+	if err := t.ensureResolved(false); err != nil {
 		return nil, err
 	}
 
 	if err := t.bspPkg.Reload(t.res.Cfg.SettingValues()); err != nil {
 		return nil, err
 	}
+
+	t.res.Cfg.DetectErrors(bspPkg.FlashMap)
 
 	return t, nil
 }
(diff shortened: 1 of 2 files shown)


kasjer:
I tested this on original blackpill411ce bsp.
With this change it builds when
```yml
syscfg.vals:
    SPIFLASH8M: 1
```
but does not when default flash map is supposed to be used
```yml
syscfg.vals:
    SPIFLASH8M: 0
```
Interestingly enough it does not build without this change at all regardless of syscfg value. (Which is strange, I think one of the versions built before).

m-gorecki:
I assume you get error "Setting NFFS_FLASH_AREA specifies unknown flash area: FLASH_AREA_NFFS".
If you are building the same target (which probably uses fs/nffs package) with just SPIFLASH8M disabled then this is expected behavior. blackpill411ce bsp sets syscfg:
`NFFS_FLASH_AREA: FLASH_AREA_NFFS
`
 and if FLASH_AREA_NFFS is not defined in the chosen flash map then it's an error and default flash map indeed does not define this area. It probably worked before because newt ignores set but undefined syscfgs. So in some other target, which did not use fs/nffs package this setting was ignored and the default flash map was fine.
This also explains why target doesn't build without this change regardless of syscfg value - without this PR default map was always used when detecting flash configuration errors.

kasjer:
@m-gorecki thanks that was it, it works as expected.
I was not aware of `type: flash_owner` in **syscfg.defs:**
```yml
syscfg.defs:
    NFFS_FLASH_AREA:
        description: 'Flash area to use for NFFS.'
        type: flash_owner
        value:
        restrictions:
            - $notnull
```

kasjer (approved):

PR #604, opened 2025-08-29
newt: Add constant syscfg for each package

m-gorecki:
Now every package included in build will add
syscfg PKG_<pkg.name> and set it to 1.
This config can be used to conditionally include
other configs or packages if some package is included.


kasjer (approved):
LGTM, tested ok

sjanc (approved):

